PR #2, opened 2022-08-15
rocThrust Example Suite

mfep:
- device_ptr: Simple program that showcases the usage of the `thrust::device_ptr` template.
- norm: An example that computes the Euclidean norm of a `thrust::device_vector`.
- reduce_sum: An example that computes the sum of a `thrust::device_vector` integer vector using the `thrust::reduce()` generalized summation and the `thrust::plus` operator.
- remove_points: Simple program that demonstrates the usage of the `thrust` random number generation, host vector, generation, tuple, zip iterator, and conditional removal templates. It generates a number of random points in a unit square and then removes all of them outside the unit circle.
- saxpy: Simple program that implements the SAXPY operation (`Y[i] = a * X[i] + Y[i]`) using rocThrust and showcases the usage of the vector and functor templates and of `thrust::fill` and `thrust::transform` operations.
- vectors: Simple program that showcases the `host_vector` and the `device_vector` of rocThrust.


Libraries/rocThrust/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Summary
+The examples in this subdirectory showcase the functionality of the [rocThrust](https://github.com/rocmSoftwarePlatform/rocThrust) library. The examples build on Linux using the ROCm platform and on Windows using the HIP on Windows platform.
+
+## Prerequisites
+### Linux
+- [CMake](https://cmake.org/download/) (at least version 3.21)
+- OR GNU Make - available via the distribution's package manager
+- [ROCm](https://docs.amd.com/bundle/ROCm-Installation-Guide-v5.2/page/Overview_of_ROCm_Installation_Methods.html) (at least version 5.x.x)
+- [rocThrust](https://github.com/rocmSoftwarePlatform/rocThrust): `rocthrust-dev` package available from [repo.radeon.com](https://repo.radeon.com/rocm/). The repository is added during the standard ROCm [install procedure](https://docs.amd.com/bundle/ROCm-Installation-Guide-v5.2/page/How_to_Install_ROCm.html).
```
saadrahim, line 11: I find using absolute links an unnecessary maintenance burden. But I guess we have no choice in the interim

Libraries/example_library_template/example_template/README.md
```diff
@@ -0,0 +1,21 @@
...
+### Application flow 
+*Explain the application flow of the example program. How the interactivity of the example is handled (e.g. via a CLI), what operations are performed by the program, and which error handling measures are in place.*
+
+### Command line interface
+*Explain the command line interface of the example program, if applicable.*
+
+## Key APIs and Concepts
+*List and explain the APIs used in the example and the library's high-level concepts behind these APIs. Give an overview of the structure of the library and how the showcased APIs can interact with other facilities of the library.*
+
+## Used API surface
```
doctorcolinsmith: Fix the heading (as per other PR)

Libraries/rocThrust/device_ptr/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+cmake_minimum_required(VERSION 3.21 FATAL_ERROR)
```
doctorcolinsmith, line 23: Is there a way to have the requirements shared across the examples?  That is it seems that all the examples need CMake 3.21, C++ 17, &c.

Libraries/rocThrust/device_ptr/Makefile
```diff
@@ -0,0 +1,34 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+ROCM_INSTALL_DIR = /opt/rocm
+HIPCXX = $(ROCM_INSTALL_DIR)/bin/hipcc
+CXX_STD = c++17
```
doctorcolinsmith, line 25: Again, it seems that there should be a way to share the above dependencies with all examples.

doctorcolinsmith (commented):
A few minor comments but overall fine.  Please also get approval from @saadrahim and @stanleytsang-amd.

Libraries/example_library_template/example_template/README.md
mfep: Done as requested.

Libraries/rocThrust/device_ptr/CMakeLists.txt
mfep: Yes, there currently is quite some duplication between the build scripts among the examples. This is a deliberate choice and serves a dual purpose:
- Each example can be copied out-of-tree and work with minimal modifications (only the utility header has to be copied with it),
- The build scripts serve as part of the tutorial itself, as (hopefully) they are easy to follow.

We thought that in the case of the build scripts the benefits above justify the duplication. If you think otherwise, we can introduce a "build script utility library", which would reduce much of it -- in that case I think not only the C++ standard and CMake version requirement should be shared, but also a lot of the logic that is required in the build scripts. A good example for that is the [CMakeLists.txt of the rocBLAS-examples](https://github.com/ROCmSoftwarePlatform/rocBLAS-Examples/blob/develop/CMakeLists.txt) which generates all CMake projects from the tree.

Please let us know if we should make this change in build scripts "philosophy" and we'll update the PRs accordingly.

Libraries/rocThrust/device_ptr/Makefile
mfep: Please see [reply here](https://github.com/amd/rocm-examples/pull/2#discussion_r955756959).

Libraries/rocThrust/device_ptr/CMakeLists.txt
doctorcolinsmith: The response makes a good point.  Leave the code as-is.

saadrahim (approved):

PR #6, opened 2022-09-12
HIP "Cookbook" Example Suite (part 1)

mfep:
- device_query: Program that showcases how properties from the device may be queried.
- dynamic_shared: Program that showcases how to use dynamic shared memory with the help of a simple matrix transpose kernel.
- hello_world: Simple program that showcases launching kernels and printing from the device.
- hipify: Simple program and build definitions that showcase automatically converting a CUDA `.cu` source into portable HIP `.hip` source.
- matrix_multiplication: Multiply two dynamically sized matrices utilizing shared memory.
- occupancy: Shows how to find optimal configuation parameters for a kernel launch with maximum occupancy.
- saxpy: Implements the $Y_i=aX_i+Y_i$ kernel and explains basic HIP functionality.


saadrahim:
I don't like the use of the term cookbook. Perhaps hipruntime instead.

Cookbook/hipify/main.cu
```diff
@@ -0,0 +1,115 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#include <cuda_runtime.h>
+
+#include <cstdlib>
+#include <iostream>
+#include <vector>
+
+#define CHECK(cmd)                                                                          \
```
saadrahim: should the checks be in common?

saadrahim: Sorry, this is a CUDA file, makes sense as is.

saadrahim (changes requested):
Please change the cookbook folder to HIP_Basic or something similar.

mfep:
> Please change the cookbook folder to HIP_Basic or something similar.

Done as requested.

saadrahim (approved):

PR #24, opened 2023-02-09
[fix][build][vcxproj][cleanup] Get rid of `HIPPropertiesImported` and `HIPTargetsImported` conditions

emankov:
**[Reason]**
+ `AMD.HIP.Clang|Nvcc.Common.props` and `AMD.HIP.Clang|Nvcc.Common.targets` are loaded unconditionally now by HIP-VS Extension
+ The corresponding changes are already implemented in HIP-VS Extension and corresponding project Templates

**[IMP]**
+ The removed conditions won't appear in `vcxproj` files of newly created (by `Create new Project` Wizard) HIP projects anymore
+ The existing projects still need cleanup


hchevanne (approved):
LGTM.
I checked that all 51 projects build correctly with VS2017 (15.9.51), VS2019 (16.11.23) and VS2022 (17.4.4)

PR #25, opened 2023-02-11
[vcxproj][fix] Use `LatestTargetPlatformVersion` as `WindowsTargetPlatformVersion` instead of `10.0` in all vcxproj files

emankov:
**[Reasons]**
+ `$(LatestTargetPlatformVersion)` is already a default value for both Platform Toolsets `HIP_clang` and `HIP_nvcc`
+ Better to have the latest Windows SDK version for the already created projects when Windows SDK is updated
+ Backward compatibility: Visual Studio 2017 users, for instance, may have Windows SDK < 10.0


hchevanne (approved):
LGTM

PR #26, opened 2023-02-12
[docker][git] Minor `Dockerfile` and `.gitignore` enhancements 

guilt:
I thank you for reviewing this PR.

The purpose of the PR is to enable an optional (group ID) `GID` argument to be passed so that it matches the
one on the host. Often, the `render` group id, when matched with the host, allows us to build and
run the examples without requiring root permissions. With that, I am able to run this on most distributions
and I can pass `--build-arg GID=$(getent group render  | cut -d: -f3)` while building with Docker.

The other fix adds the `Build` directory to `.gitignore`, in-line with the naming convention.

Commit Message:

```
Build Fixes for various distributions.

* Change groupid to an optional argument a user can pass when building.
* Add Build to .gitignore
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@
 [Dd]ebug
 [Rr]elease
 *.vcxproj.user
-build
+[Bb]uild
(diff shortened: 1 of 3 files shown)


saadrahim:
I think this looks good. Can someone approve it?

emankov (approved):
LGTM, but I'd say that it is Docker / .gitignore minor enhancements, not build fixes

PR #27, opened 2023-02-15
[vcxproj][nvcc] Exclude 27 projects from building by `HIP_nvcc` toolset

emankov:
**[Reason]**
+ Not ready for the nvcc path

**[IMP]**
+ The excluded projects won't be built, the corresponding `warning` will be emitted:
image


saadrahim (approved):
LGTM

PR #28, opened 2023-02-17
[nvcc][fix] Fix compilation problems with HIP_nvcc toolchain

hchevanne:
**Fix several compilation problems with HIP_nvcc toolchain (CUDA Toolkit 11.8):**

- HIP-Basic/matrix_multiplication
> main.hip
- HIP-Basic/moving_average
> main.hip
> moving_average_vs2017.vcxproj
> moving_average_vs2019.vcxproj
> moving_average_vs2022.vcxproj
- HIP-Basic/multi_gpu_data_transfer
> main.hip
- HIP-Basic/opengl_interop
> opengl_interop_vs2022.vcxproj
- HIP-Basic/shared_memory
> main.hip
- HIP-Basic/streams
> streams_vs2017.vcxproj
> streams_vs2019.vcxproj
> streams_vs2022.vcxproj
- HIP-Basic/warp_shuffle
> main.hip

### DIFF
--- a/HIP-Basic/matrix_multiplication/main.hip
+++ b/HIP-Basic/matrix_multiplication/main.hip
@@ -184,7 +184,7 @@ int main(int argc, const char* argv[])
     HIP_CHECK(hipMemcpy(d_A, A.data(), a_bytes, hipMemcpyHostToDevice));
     HIP_CHECK(hipMemcpy(d_B, B.data(), b_bytes, hipMemcpyHostToDevice));
 
-    constexpr dim3 block_dim(block_size, block_size);
+    const dim3 block_dim(block_size, block_size);
     const dim3     grid_dim(c_cols / block_size, c_rows / block_size);
 
     // Launch matrix multiplication kernel.
--- a/HIP-Basic/moving_average/main.hip
+++ b/HIP-Basic/moving_average/main.hip
@@ -101,7 +101,7 @@ int main()
     std::transform(h_input.begin(),
                    h_input.end(),
                    h_input.begin(),
-                   [](unsigned int i) { return i % window_size; });
+                   [&](unsigned int i) { return i % window_size; });
 
     // Allocate device input data and copy host data to it.
     unsigned int*    d_input{};
--- a/HIP-Basic/moving_average/moving_average_vs2017.vcxproj
+++ b/HIP-Basic/moving_average/moving_average_vs2017.vcxproj
@@ -65,6 +65,12 @@
   <PropertyGroup Label="HIP_clang" Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
     <OffloadArch>gfx1030;gfx1100;gfx1101;gfx1102</OffloadArch>
   </PropertyGroup>
+  <PropertyGroup Label="HIP" Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <NvccAdditionalOptions>-std=c++17</NvccAdditionalOptions>
+  </PropertyGroup>
+  <PropertyGroup Label="HIP" Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <NvccAdditionalOptions>-std=c++17</NvccAdditionalOptions>
+  </PropertyGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64' and '$(PlatformToolset)'=='HIP_clang'">
     <ClCompile>
       <WarningLevel>Level2</WarningLevel>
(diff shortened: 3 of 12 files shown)


HIP-Basic/matrix_multiplication/main.hip
```diff
@@ -184,7 +184,7 @@ int main(int argc, const char* argv[])
     HIP_CHECK(hipMemcpy(d_A, A.data(), a_bytes, hipMemcpyHostToDevice));
     HIP_CHECK(hipMemcpy(d_B, B.data(), b_bytes, hipMemcpyHostToDevice));
 
-    constexpr dim3 block_dim(block_size, block_size);
+    const dim3 block_dim(block_size, block_size);
```
emankov, line 187: Better to see the error reported in the description first.

HIP-Basic/moving_average/main.hip
```diff
@@ -101,7 +101,7 @@ int main()
     std::transform(h_input.begin(),
                    h_input.end(),
                    h_input.begin(),
-                   [](unsigned int i) { return i % window_size; });
+                   [&](unsigned int i) { return i % window_size; });
```
emankov, line 104: Better to see the error reported in the description first. Also, better to have fixes for different types of errors in different PRs.

HIP-Basic/moving_average/moving_average_vs2017.vcxproj
```diff
@@ -65,6 +65,12 @@
   <PropertyGroup Label="HIP_clang" Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
     <OffloadArch>gfx1030;gfx1100;gfx1101;gfx1102</OffloadArch>
   </PropertyGroup>
+  <PropertyGroup Label="HIP" Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <NvccAdditionalOptions>-std=c++17</NvccAdditionalOptions>
+  </PropertyGroup>
+  <PropertyGroup Label="HIP" Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <NvccAdditionalOptions>-std=c++17</NvccAdditionalOptions>
+  </PropertyGroup>
```
emankov, line 73: C++ `LanguageStandard` is implemented for `nvcc` finally, so better to set it in the corresponding PPP, for instance:

```xml
  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64' and '$(PlatformToolset)'=='HIP_nvcc'">
    <ClCompile>
      <PreprocessorDefinitions>__CUDACC__;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
      <LanguageStandard>stdcpp17</LanguageStandard>
    </ClCompile>
  </ItemDefinitionGroup> 
```
Btw, what about `HIP_clang`, is it ok with the default `c++14`?

emankov (changes requested):

HIP-Basic/moving_average/moving_average_vs2017.vcxproj
hchevanne: Of course.
Compilation errors fixed by this PR are all related to host code. Unlike Clang, in the case of compiling host code, nvcc uses the system's native compiler: Microsoft cl in this case. 
Using the c++17 standard is required by cl but not by Clang.

emankov: Thanks, but 1) better to have an error in the PR description as well 2) it's a bit strange that for `clang` it is enough to have c++14, and for `cl` is not; sounds like some `c++17` feature is introduced in `c++14` for clang - if it is so, it is a very unusual situation.

Naraenda: This stems from [the call to the single argumented version of `static_assert()`](https://github.com/amd/rocm-examples/blob/ff73ec0937dfa5f912ee9d916d58ba0bac628230/HIP-Basic/moving_average/main.hip#L87) which is introduced in c++17. Clang CL gives a warning while MSVC CL 19.29 (bundled with VS2019) raises an error. [See Godbolt.](https://godbolt.org/z/4rbhqvPxq) Maybe a better solution is to use the two argument variant of static_assert?

emankov: @saadrahim, can HIP sources be touched in order to fix the `nvcc` path only?

Snektron:
Whats the status on this? It would be great to get these changes, so that we don't have to do any duplicate work.

HIP-Basic/moving_average/moving_average_vs2017.vcxproj
Snektron: Why not just change the language standard to c++17 for HIP_clang as well?

hchevanne:
I checked with VS plugin build #94. This PR in its entirety is no longer required.
I close it to avoid polluting the project files involved.
New PR will eventually be necessary if problems persist with nvcc toolchain.

PR #29, opened 2023-02-27
Visual Studio improvements (II)

Snektron:
* Improves copying runtime compilation dependencies
  * Also improves the way that they are copied
* Disables OpenGL example on Windows


hchevanne (approved):
LGTM

PR #32, opened 2023-02-28
Regenerate assembly-to-executable source code for GFX11

Snektron:
This pull request updates the assembly-to-executable sources for GFX11. Note, these do not compile with any current version of ROCm clang. Please test whether this runs on GFX11 when compiled with a sufficiently new version of clang, where this patch has landed: https://reviews.llvm.org/D134272. I previously compiled successfully with commit a74274e615d5af34867976c36845104ade14e084 from rocm-llvm (https://github.com/RadeonOpenCompute/llvm-project/commit/a74274e615d5af34867976c36845104ade14e084).

Steps to test:
1. Clone this pull request
2. Compile above version of rocm-clang. Note: any relatively recent version will probably also do.
3. Navigate to `HIP-Basic/assembly-to-executable`
4. Run `make CLANG=path/to/custom/llvm/bin/clang++`
5. If compilation is successful, run `hip_assembly_to_executable` on a GFX11 device.

### DIFF
--- a/HIP-Basic/assembly_to_executable/main_gfx1100.s
+++ b/HIP-Basic/assembly_to_executable/main_gfx1100.s
@@ -12,15 +12,15 @@ _Z20vector_square_kernelIfEvPT_PKS0_y:  ; @_Z20vector_square_kernelIfEvPT_PKS0_y
 	s_waitcnt lgkmcnt(0)
 	s_and_b32 s4, s4, 0xffff
 	s_delay_alu instid0(SALU_CYCLE_1) | instskip(SKIP_2) | instid1(VALU_DEP_1)
-	v_mad_u64_u32 v[2:3], null, s15, s4, v[0:1]
+	v_mad_u64_u32 v[1:2], null, s15, s4, v[0:1]
 	v_mov_b32_e32 v2, 0
 	s_mov_b32 s4, exec_lo
-	v_cmpx_gt_u64_e64 s[8:9], v[2:3]
+	v_cmpx_gt_u64_e64 s[8:9], v[1:2]
 	s_cbranch_execz .LBB0_3
 ; %bb.1:
 	s_load_b32 s10, s[0:1], 0xc
 	s_load_b128 s[4:7], s[2:3], 0x0
-	v_lshlrev_b64 v[4:5], 2, v[2:3]
+	v_lshlrev_b64 v[3:4], 2, v[1:2]
 	s_mov_b32 s11, 0
 	s_waitcnt lgkmcnt(0)
 	s_lshl_b64 s[2:3], s[10:11], 2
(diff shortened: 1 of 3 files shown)


Snektron:
Please dont merge until above compiler version is in ROCm.

Snektron:
Added an extra commit to work around the required clang version. Please test again, but with a regular ROCm 5.3.x or 5.4.x build of clang.

saadrahim:
@Snektron are we ready to merge this?

Snektron:
Yes, the test passes on our setup.

Snektron:
@saadrahim i cannot merge this without review..

emankov:
@Snektron, Is my understanding correct that with merging #36 we do not need this PR at all and it can be just closed?

emankov (commented):
LGTM, otherwise it is better to merge #36 only.

Snektron:
> @Snektron, Is my understanding correct that with merging #36 we do not need this PR at all and it can be just closed?

That is a different sample

emankov:
> Added an extra commit to work around the required clang version. Please test again, but with a regular ROCm 5.3.x or 5.4.x build of clang.

What about ROCm 5.5 we are targeting? I believe we shouldn't care about anything pre-5.5 on Windows.

Snektron:
> > Added an extra commit to work around the required clang version. Please test again, but with a regular ROCm 5.3.x or 5.4.x build of clang.
> 
> What about ROCm 5.5 we are targeting? I believe we shouldn't care about anything pre-5.5 on Windows.

This current version should work for ROCm 5.5 as well. The original issue was that pre-5.5 compilers could not parse the assembly generated by clang, even though it is correct, so it was manually modified so that clang could parse it. At the time we were not able to test that due to a lack of gfx11 hardware, and hence this PR which fixes some mistakes that were made back then. In theory, for ROCm 5.5 this manual edit is no longer required as the commit that fixes this issue in LLVM made its way into ROCm, and we can regenerate the assembly as described in the README of this example. Do you think that is preferred?

emankov:
> > > Added an extra commit to work around the required clang version. Please test again, but with a regular ROCm 5.3.x or 5.4.x build of clang.
> > 
> > 
> > What about ROCm 5.5 we are targeting? I believe we shouldn't care about anything pre-5.5 on Windows.
> 
> This current version should work for ROCm 5.5 as well. The original issue was that pre-5.5 compilers could not parse the assembly generated by clang, even though it is correct, so it was manually modified so that clang could parse it. At the time we were not able to test that due to a lack of gfx11 hardware, and hence this PR which fixes some mistakes that were made back then. In theory, for ROCm 5.5 this manual edit is no longer required as the commit that fixes this issue in LLVM made its way into ROCm, and we can regenerate the assembly as described in the README of this example. Do you think that is preferred?

If I understood correctly, the `assembly-to-executable` should be tested without the #32 fix against the latest available ROCm 5.5. And if everything is fine, there is no need to merge #32, only assembly regeneration and running check on the gfx11 are needed.

saadrahim (approved):

emankov:
@saadrahim, could you please explain why you decided to merge this change? There was some discussion about checking against the latest 5.5 without this change.

saadrahim:
It was my understanding that both PRs were needed. Also, merging this will allow QA to test it. Is there a need to revert?

emankov:
> It was my understanding that both PRs were needed. Also, merging this will allow QA to test it. Is there a need to revert?

As for this PR, the testing of `assembly-to-executable` against the latest 5.5 on gfx11 is needed, but **without this PR** :) 
So, my understanding is that probably we do not need this PR if everything is fine on 5.5 + gfx11.

Snektron:
> So, my understanding is that probably we do not need this PR if everything is fine on 5.5 + gfx11.

To be clear, a PR would still need to be made regardless. The assembly that was previously used for this example was definitely wrong. There were basically 2 options:
- Keep it like this. This assembly is the generated one except with a few additions to make it able to be compiled on HIP 5.3 and 5.4.
- Regenerate the assembly, after which it will only work with 5.5 and above.

Personally I think its nice that the example works with ROCm 5.3 and 5.4, though I don't feel particularly strong about that.

emankov:
> Personally I think its nice that the example works with ROCm 5.3 and 5.4, though I don't feel particularly strong about that.

There aren't any 5.3 or 5.4, and won't ever be. The first HIP-SDK for Windows will be 5.5. 

So, I'm strongly for regenerating the assembly for 5.5; the same should be accomplished for 5.6 if needed.

Colleagues,
please, give exact instructions for QA to test this example on ROCm 5.5 RC + gfx11 without this change first. Based on testing results, we'll decide whether to rollback this change or leave it for 5.5.

Thanks!

Snektron:
> There aren't any 5.3 or 5.4, and won't ever be. The first HIP-SDK for Windows will be 5.5.

But there is for Linux, and the examples in this repository are also for that.

> Colleagues,
please, give exact instructions for QA to test this example on ROCm 5.5 RC + gfx11 without this change first. Based on testing results, we'll decide whether to rollback this change or leave it for 5.5.

It should be sufficient to check out a version of this repository from before the merge (for example fc00d4643a735043d6fca5d3b00b73ddc737046e) and build + run the example on a machine with a gfx11 card.

emankov:
> But there is for Linux, and the examples in this repository are also for that.

There were no releases of rocm-examples yet. As for Linux, the latest ROCm release is 5.5 now. It means, in the first release of rocm-examples the minimum supported ROCm version will be at least >= 5.5. So, we do not need to care about the pre-5.5 compatibility of the rocm-examples on any OS.

PR #37, opened 2023-05-24
rocSOLVER and hipSOLVER examples (part II)

Beanavil:
This pull request completes the rocSOLVER batch of examples and adds some hipSOLVER examples. A third (and last) PR will be opened once this one is merged, adding the remaining hipSOLVER examples.

### New examples
- hipSOLVER
  - getrf
 
- rocSOLVER
  - getf2
  - syev
  - syev_batched
  - syev_strided_batched

### Renamed examples
- rocSOLVER 
  | Old name          | New name |
  |-------------------|----------|
  |`matrix_inversion` | `getri`  |

### New directories order

```
├── hipSOLVER
│   ├── cholesky_decomposition
│   ├── generalized_symmetric_eigenvalue_divide_and_conquer
│   ├── getrf
│   ├── linear_least_squares
│   ├── qr_decomposition
│   ├── singular_value_decomposition
│   ├── symmetric_eigenvalue_divide_and_conquer
.
.
.
├── rocSOLVER
│   ├── getf2
│   ├── getri
│   ├── syev
│   ├── syev_batched
│   └── syev_strided_batched
```


Beanavil:
cc @tfalders

Libraries/hipSOLVER/getrf/README.md
```diff
@@ -0,0 +1,68 @@
+# hipSOLVER LU Factorization Example
+
+## Description
+This example illustrates the use of the hipSOLVER LU factorization functionality. The hipSOLVER `getrf` computes the [LU decomposition](https://en.wikipedia.org/wiki/LU_decomposition) of an $m \times n$ matrix $A$, with partial pivoting. This factorization is given by $P \cdot A = L \cdot U$, where:
+- `getrf()`: This is the blocked Level-3-BLAS version of the LU factorization algorithm. An optimized internal implementation without rocBLAS calls could be executed with mid-size matrices.
+- $A$ is the $m \times n$ input matrix.
+- $P$ is an $m \times m$ [permutation matrix](https://en.wikipedia.org/wiki/Permutation_matrix), in this example stored as column representation vector in `vector<int> Ipiv` of size `min(m, n)`.
+- $L$ is an $m \times m$ lower triangular matrix.
+- $U$ is an $m \times n$ upper triangular matrix.
```
tfalders: I believe the dimensions of these matrices depend on whether `m > n` or `m < n`. For instance, I'm pretty sure L is `m*m` when `m < n`, but it's `m*n` when `m < n`.

Libraries/hipSOLVER/getrf/README.md
```diff
@@ -0,0 +1,68 @@
...
+### Application flow
+1. Parse command line arguments for the dimension of the input matrix.
+2. Declare and initialize a number of constants for the input and output matrices and vectors.
+3. Allocate and initialize the host matrices and vectors.
+4. Allocate device memory.
+5. Copy input data from host to device.
+6. Create a `hipsolverHandle_t` handle.
+7. Invoke the hipSOLVER `getrf` function with double precision.
+8. Copy the results from device to host.
+9. Print the LU matrix and Ipiv vector.
```
tfalders: Based on my reading of the example, I don't believe it prints these.

Libraries/hipSOLVER/getrf/README.md
```diff
@@ -0,0 +1,68 @@
+# hipSOLVER LU Factorization Example
+
+## Description
+This example illustrates the use of the hipSOLVER LU factorization functionality. The hipSOLVER `getrf` computes the [LU decomposition](https://en.wikipedia.org/wiki/LU_decomposition) of an $m \times n$ matrix $A$, with partial pivoting. This factorization is given by $P \cdot A = L \cdot U$, where:
+- `getrf()`: This is the blocked Level-3-BLAS version of the LU factorization algorithm. An optimized internal implementation without rocBLAS calls could be executed with mid-size matrices.
+- $A$ is the $m \times n$ input matrix.
+- $P$ is an $m \times m$ [permutation matrix](https://en.wikipedia.org/wiki/Permutation_matrix), in this example stored as column representation vector in `vector<int> Ipiv` of size `min(m, n)`.
```
tfalders: "Column representation vector" doesn't sound quite right to me. I think it would be more accurate to call it an "array of row indices".

Libraries/rocSOLVER/getf2/README.md
```diff
@@ -0,0 +1,59 @@
+# rocSOLVER LU Factorization Example
+
+## Description
+This example illustrates the use of the rocSOLVER `getf2` functionality. The rocSOLVER `getf2` computes the [LU decomposition](https://en.wikipedia.org/wiki/LU_decomposition) of an $m \times n$ matrix $A$, with partial pivoting. This factorization is given by $P \cdot A = L \cdot U$, where:
+- `getf2()`: This is the unblocked Level-2-BLAS version of the LU factorization algorithm. An optimized internal implementation without rocBLAS calls could be executed with small and mid-size matrices.
+- $A$ is the $m \times n$ input matrix.
+- $P$ is an $m \times m$ [permutation matrix](https://en.wikipedia.org/wiki/Permutation_matrix), in this example stored as column representation vector in `vector<int> Ipiv` of size `min(m, n)`.
+- $L$ is an $m \times m$ lower triangular matrix.
+- $U$ is an $m \times n$ upper triangular matrix.
```
tfalders: Same comments as for hipsolver_getrf

Libraries/rocSOLVER/getf2/README.md
```diff
@@ -0,0 +1,59 @@
...
+### Application flow
+1. Parse command line arguments for the dimension of the input matrix.
+2. Declare and initialize a number of constants for the input and output matrices and vectors.
+3. Allocate and initialize the host matrices and vectors.
+4. Allocate device memory.
+5. Copy input data from host to device.
+6. Create a rocBLAS handle.
+7. Invoke the rocSOLVER `getf2` function with double precision.
+8. Copy the results from device to host.
+9. Print the LU matrix and Ipiv vector.
```
tfalders: Again, I don't believe these are being printed.

Libraries/rocSOLVER/syev/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Key APIs and Concepts
+
+### rocSOLVER
+
+- `rocblas_create_handle(...)` initializes rocblas.
+- `rocsolver_[sd]syev(...)` computes the eigenvalue and optional eigenvector.
+  - There are 2 different function signatures depending on the type of the input matrix:
+    - `s` single-precision real (`float`)
+    - `d` double-precision real (`double`)
```
tfalders, line 42: It might be worth mentioning that rocsolver_[cz]heev is the complex-valued version of the function

Libraries/rocSOLVER/syev/main.cpp
```diff
@@ -0,0 +1,167 @@
...
+    std::uniform_real_distribution<rocblas_double> distribution(0., 2.);
+    auto random_number = [&]() { return distribution(generator); };
+
+    for(int i = 0; i < n; i++)
+    {
+        A[(n + 1) * i] = random_number();
+        for(int j = 0; j < i; j++)
+        {
+            A[i * n + j] = A[j * n + i] = random_number();
+        }
```
tfalders: Using lda instead of n here would be safer, e.g. in case a user changed the value of lda for testing purposes. Same goes for the multiply_matrices calls below.

Libraries/rocSOLVER/syev/README.md
```diff
@@ -0,0 +1,73 @@
...
+### rocSOLVER
+
+- `rocsolver_dsyev`
+
+### rocBLAS
+
+- `rocblas_create_handle`
+- `rocblas_destroy_handle`
+- `rocblas_double`
+- `rocblas_evect::rocblas_evect_original`
```
tfalders: `rocblas_evect` is actually part of the rocSOLVER API, which is a bit confusing, but we wanted the enums to have a consistent naming scheme.

Libraries/rocSOLVER/syev/main.cpp
```diff
@@ -0,0 +1,167 @@
...
+    HIP_CHECK(hipMalloc(&d_A, sizeof(rocblas_double) * A.size()));
+    HIP_CHECK(hipMalloc(&d_W, sizeof(rocblas_double) * W.size()));
+    HIP_CHECK(hipMemcpy(d_A, A.data(), sizeof(rocblas_double) * A.size(), hipMemcpyHostToDevice));
+
+    // 6. Initialize rocBLAS.
+    rocblas_handle handle;
+    ROCBLAS_CHECK(rocblas_create_handle(&handle));
+
+    // 7. Get and reserve the working space on device.
+    rocblas_int     d_work_size = n * 3 - 1;
```
tfalders: I believe this should be size n.

Libraries/rocSOLVER/syev_strided_batched/README.md
```diff
@@ -0,0 +1,133 @@
...
+The solver evaluates the following equation for a strided batch of $m$ symmetric matrices, named as $A_i$, and with the size of $n \times n$:
+
+$$A_i \cdot V_i = W_i \cdot V_i$$
+
+for each $0 \leq i < m$.
+
+The set of orthonormalized eigenvectors can be settled to a column of a matrix as
+
+$$
+V_i = \left[\mathbf{x_{i_0}}, \dots, \mathbf{x_{i_j}}, \dots, \mathbf{x_{i_{n-1}}}\right]
```
tfalders, line 24: Is there a reason why these are x_{i_0}, x_{i_j}, etc... instead of x_0, x_j, etc...?

Libraries/rocSOLVER/syev_strided_batched/README.md
```diff
@@ -0,0 +1,133 @@
...
+9. Retrieve the results by copying from device to host.
+10. Print the results
+11. Validate the results
+12. Free the memory allocations on device.
+
+## Key APIs and Concepts
+- The performance of a numerical multi-linear algebra code can be heavily increased by using tensor contractions [ [Y. Shi et al., HiPC, pp 193, 2016.](https://doi.org/10.1109/HiPC.2016.031) ], thereby similarly to other linear algebra libraries like hipBLAS rocSOLVER also has a `_batched` and a `_strided_batched` [ [C. Jhurani and P. Mullowney, JPDP Vol 75, pp 133, 2015.](https://doi.org/10.1016/j.jpdc.2014.09.003) ] extensions.<br/>
+We can apply the same operation for several matrices if we combine them into batched matrices. Batched computation has a performance improvement for a large number of small matrices. For a constant stride between matrices, further acceleration is available by strided batched solvers.
+
+### rocSOLVER
```
tfalders, line 80: Same comments as for syev

Libraries/rocSOLVER/syev_strided_batched/main.cpp
```diff
@@ -0,0 +1,286 @@
...
+#include <hip/hip_runtime.h>
+
+#include <iomanip>
+#include <iostream>
+#include <numeric>
+#include <random>
+#include <string>
+#include <vector>
+
+int main(const int argc, char* argv[])
```
tfalders, line 39: Same comments as for syev

Libraries/rocSOLVER/syev_strided_batched/main.cpp
```diff
@@ -0,0 +1,286 @@
...
+                                                  evect,
+                                                  uplo,
+                                                  n,
+                                                  d_A,
+                                                  lda,
+                                                  single_matrix_num_elements_padded,
+                                                  d_W,
+                                                  single_eigenvalue_num_elements_padded,
+                                                  d_work,
+                                                  n * n,
```
tfalders: Shouldn't this be batch_tridiagonal_num_elements_padded?

In fact, using `n*n` as the stride will probably segfault for most values of n, since d_work only has `n*3-1` (should be n) elements per batch and not `n*n`.

Libraries/rocSOLVER/syev_batched/README.md
```diff
@@ -0,0 +1,130 @@
...
+The solver evaluates the following equation for a strided batch of $m$ symmetric matrices, named as $A_i$, and with the size of $n \times n$:
+
+$$A_i \cdot V_i = W_i \cdot V_i$$
+
+for each $0 \leq i < m$.
+
+The set of orthonormalized eigenvectors can be settled to a column of a matrix as
+
+$$
+V_i = \left[\mathbf{x_{i_0}}, \dots, \mathbf{x_{i_j}}, \dots, \mathbf{x_{i_{n-1}}}\right]
```
tfalders, line 24: Is there a reason why these are x_{i_0}, x_{i_j}, etc... instead of x_0, x_j, etc...?

Libraries/rocSOLVER/syev_batched/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+Generally, in an eigenvalue problem, we are looking for $\mathbf{x}$ vectors with $\lambda$ scalars that fulfill the
+
+$$
+A_i \cdot \mathbf{x} = \lambda \cdot \mathbf{x}
+$$
+
+equation.
+
+The solver evaluates the following equation for a strided batch of $m$ symmetric matrices, named as $A_i$, and with the size of $n \times n$:
```
tfalders: For syev_batched, the matrices are not a strided batch

Libraries/rocSOLVER/syev_batched/README.md
```diff
@@ -0,0 +1,130 @@
...
+9. Retrieve the results by copying from device to host.
+10. Free the memory allocations on device.
+11. Print the results
+12. Validate the results
+
+## Key APIs and Concepts
+- The performance of a numerical multi-linear algebra code can be heavily increased by using tensor contractions [ [Y. Shi et al., HiPC, pp 193, 2016.](https://doi.org/10.1109/HiPC.2016.031) ], thereby similarly to other linear algebra libraries like hipBLAS rocSOLVER also has a `_batched` and a `_strided_batched` [ [C. Jhurani and P. Mullowney, JPDP Vol 75, pp 133, 2015.](https://doi.org/10.1016/j.jpdc.2014.09.003) ] extensions.<br/>
+We can apply the same operation for several matrices if we combine them into batched matrices. Batched computation has a performance improvement for a large number of small matrices. For a constant stride between matrices, further acceleration is available by strided batched solvers.
+
+### rocSOLVER
```
tfalders, line 78: Same comments as for syev

Libraries/rocSOLVER/syev_batched/main.cpp
```diff
@@ -0,0 +1,264 @@
...
+#include <hip/hip_runtime.h>
+
+#include <iomanip>
+#include <iostream>
+#include <numeric>
+#include <random>
+#include <string>
+#include <vector>
+
+int main(const int argc, char* argv[])
```
tfalders, line 39: Same comments as for syev

Libraries/rocSOLVER/syev_batched/main.cpp
```diff
@@ -0,0 +1,264 @@
...
+    ROCBLAS_CHECK(rocsolver_dsyev_batched(handle,
+                                          evect,
+                                          uplo,
+                                          n,
+                                          array_dA,
+                                          lda,
+                                          d_W,
+                                          batch_size,
+                                          d_work,
+                                          n * n,
```
tfalders: This should be n.

Using `n*n` as the stride will probably segfault when n>1 and batch_size>1, since d_work only has n elements per batch and not `n*n`.

Libraries/rocSOLVER/syev_batched/main.cpp
```diff
@@ -0,0 +1,264 @@
...
+            std::cout << "SYEV does not converge (" << info[batch_id]
+                      << " elements did not converge)." << std::endl;
+        }
+
+        std::cout << "\nGiven the n x n square input matrix A; we computed the orthonormal "
+                     "eigenvectors V and the associated eigenvalues W."
+                  << std::endl;
+        auto A_begin = A.begin();
+        auto W_begin = W.begin();
+        auto V_begin = V.begin();
```
tfalders: This is only valid for the first batch instance (i.e. batch_id = 0)

Libraries/rocSOLVER/syev_batched/main.cpp
```diff
@@ -0,0 +1,264 @@
...
+    rocblas_double*  array_A[batch_size];
+    rocblas_double** array_dA;
+    rocblas_double*  d_A    = nullptr;
+    rocblas_double*  d_W    = nullptr;
+    rocblas_int*     d_info = nullptr;
+
+    const rocblas_int info_size  = sizeof(rocblas_int) * batch_size;
+    const rocblas_int array_size = batch_size * sizeof(rocblas_double**);
+
+    const rocblas_int full_size = batch_size * single_matrix_num_elements * sizeof(rocblas_double*);
```
tfalders: Should that be `sizeof(rocblas_double)` rather than `sizeof(rocblas_double*)`?

tfalders (commented):
Looks good! I have a few comments to be addressed before I give my approval, but thanks for taking care of this.

hchevanne:
From a Visual Studio perspective, the PropertyGroup "**_ProjectExcludedFromBuild_**" must be set to "**_true_**" in the following vcxproj files:

    └── Libraries
        ├── hipSOLVER
        │   ├── getrf
        │   │   ├── getrf_vs2017.vcxproj
        │   │   ├── getrf_vs2019.vcxproj
        │   │   └── getrf_vs2022.vcxproj
        └── rocSOLVER
            ├── getf2
            │   ├── getf2_vs2017.vcxproj
            │   ├── getf2_vs2019.vcxproj
            │   └── getf2_vs2022.vcxproj
            └── syev_strided_batched
                ├── syev_strided_batched_vs2017.vcxproj
                ├── syev_strided_batched_vs2019.vcxproj
                └── syev_strided_batched_vs2022.vcxproj

Libraries/hipSOLVER/getrf/README.md
Beanavil: You're right! I modified the description so it's clear which sizes they have in each case.

Libraries/hipSOLVER/getrf/README.md
Beanavil: Good catch. Fixed!

Libraries/hipSOLVER/getrf/README.md
Beanavil: Changed!

Libraries/rocSOLVER/getf2/README.md
Beanavil: Same answer `:)`

Libraries/rocSOLVER/getf2/README.md
Beanavil: Yep, updated so it's correct.

Libraries/rocSOLVER/syev/README.md
Beanavil: Information added! Also for syev_batched and syev_strided_batched.

Libraries/rocSOLVER/syev/main.cpp
Beanavil: True, updated it!

Libraries/rocSOLVER/syev/README.md
Beanavil: I see. I moved it to the rocSOLVER section.

Libraries/rocSOLVER/syev/main.cpp
Beanavil: Actually, there is an open discussion about this https://github.com/ROCmSoftwarePlatform/rocSOLVER/issues/543. Supposedly the workspace should have enough space to store a tridiagonal matrix (used internally in the algorithm) so that is at least `n + n - 1 + n - 1` = `3 * n - 2` elements. However, as mentioned in the discussion, when [testing syev](https://github.com/ROCmSoftwarePlatform/rocSOLVER/blob/1a4498cf91c68f147efdf70868f28c4b489b8fc0/clients/include/testing_syev_heev.hpp#L302) they use `3 * n - 1` for the workspace size, so I think it's ok to use this value for now.

Libraries/rocSOLVER/syev_strided_batched/README.md
Beanavil: Each $V_i$ has columns $x_{i_0}, x_{i_1}, \dots x_{i_{n-1}}$ (which are the eigenvectors), it's just a way of identifying each column of each $V_i$. If we didn't write it like that there wouldn't be any difference between the column vectors of $V_i$ and $V_j$ e.g., notation-wise speaking.

Libraries/rocSOLVER/syev_strided_batched/README.md
Beanavil: Got it, fixed!

Libraries/rocSOLVER/syev_strided_batched/main.cpp
Beanavil: Same answers as for syev `:)`

Libraries/rocSOLVER/syev_strided_batched/main.cpp
Beanavil: Actually, it should be `single_tridiagonal_num_elements_padded`, as this argument is meant to be the stride between the workspaces in the batch of workspaces `d_work`, which has size ` batch_tridiagonal_num_elements_padded` = `single_tridiagonal_num_elements_padded * batch_size`.

Libraries/rocSOLVER/syev_batched/README.md
Beanavil: Same answer as in the previous comment https://github.com/amd/rocm-examples/pull/37#discussion_r1224299931.

Libraries/rocSOLVER/syev_batched/README.md
Beanavil: Nice catch. Fixed!

Libraries/rocSOLVER/syev_batched/README.md
Beanavil: Same answers `:)`

Libraries/rocSOLVER/syev_batched/main.cpp
Beanavil: Same answers `:)`

Libraries/rocSOLVER/syev_batched/main.cpp
Beanavil: Actually, I believe it should be `3 * n - 1` for the reasons already stated in previous comments. I defined
```
const rocblas_int single_tridiagonal_num_elements = n * 3 - 1;
const rocblas_int batch_tridiagonal_num_elements  = single_tridiagonal_num_elements * batch_size;
```
and changed the workspace size (currently it was `n * batch_size`) to `batch_tridiagonal_num_elements` and the mentioned parameter to `single_tridiagonal_num_elements`.

Also, the parameter `batch_size` passed right after `d_W` should be `n`, because it's supposed to be the stride between the eigenvalues vector for each matrix in the batch. I added a `single_eigenvalue_num_elements = n` constant and used that instead.

Libraries/rocSOLVER/syev_batched/main.cpp
Beanavil: Indeed. Fixed it!

Libraries/rocSOLVER/syev_batched/main.cpp
Beanavil: Looks like so. Changed!

Beanavil:
PropertyGroup "ProjectExcludedFromBuild" has been set to "true" in the mentioned `.vcxproj ` files @hchevanne. 

Also added `.vcxproj.filters` for the rocSOLVER/syev_strided_batched example.

Libraries/rocSOLVER/syev/main.cpp
tfalders: I should probably have myself reassigned to that issue, since it seems like it hasn't had any movement for a while.

The tridiagonal matrix is internally stored in arrays D and E (called dW and d_work here). D contains the `n` diagonal elements (which converge to the eigenvalues), and E contains the `n-1` off-diagonal elements (which converge to zero; it's a symmetric tridiagonal matrix, so only one off-diagonal is needed). E (aka d_work) of size `n` should be sufficient for the function to work; if not, then that's a bug and I'll need to take a look at it. Using `3*n-1` as the size isn't wrong, per se, but it is using more device memory than what is needed.

Beanavil: Yes, it would be great if we could close that issue. Okay so based on this new information we could as well use `n - 1` for the `d_work` size, and that is supposedly the minimum size needed. I feel like it would be best for the examples to use this value instead of `n`, because this way noob users of this library can be made aware of the fact that at least `n - 1` is needed.

PD: it would also be great if could have this information added to the documentation.

Beanavil: Changes implemented in the last commit :)

tfalders (approved):
Thank you for the reminder, @Beanavil! The PR looks good to me :)

Naraenda (approved):

PR #39, opened 2023-06-28
rocSOLVER and hipSOLVER examples (part III)

Beanavil:
This pull request completes the roc/hipSOLVER batch of examples.

### New examples
- hipSOLVER
  - syevdx
  - syevj
  - syevj_batched
  - sygvj

### Renamed examples
- hipSOLVER 

  |Old name                                             | New name |
  |-----------------------------------------------------|----------|
  |`cholesky_decomposition`                             | `potrf`  |
  |`generalized_symmetric_eigenvalue_divide_and_conquer`| `sygvd`  |
  |`linear_least_squares`                               | `gels`   |
  |`qr_decomposition`                                   | `geqrf`  |
  |`symmetric_eigenvalue_divide_and_conquer`            | `syevd`  |
  |`singular_value_decomposition`                       | `gesvd`  |

### New (and final) directories order

```
├── hipSOLVER
│   ├── gels
│   ├── geqrf
│   ├── gesvd
│   ├── getrf
│   ├── potrf
│   ├── syevd
│   ├── syevdx
│   ├── syevj
│   ├── syevj_batched
│   ├── sygvd
│   └── sygvj
.
.
.
├── rocSOLVER
│   ├── getf2
│   ├── getri
│   ├── syev
│   ├── syev_batched
│   └── syev_strided_batched
```


Beanavil:
cc @tfalders

Libraries/hipSOLVER/syevdx/README.md
```diff
@@ -0,0 +1,132 @@
...
+9. Free device memory and the handles.
+10. Print validation result.
+
+## Key APIs and Concepts
+### hipSOLVER
+- hipSOLVER is initialized by calling `hipsolverDnCreate(hipsolverHandle_t*)` and it is terminated by calling `hipsolverDnDestroy(hipsolverHandle_t)`.
+- In this example `hipsolverDnHandle_t` is used instead of `hipsolverHandle_t`. `hipsolverDnHandle_t` is actually a typedef of `hipsolverHandle_t`, so they can be used equivalently.
+- `hipsolverDn[SD]syevdx` computes the eigenvalues of an $n \times n$ symmetric matrix $A$. The correct function signature should be chosen based on the datatype of the input matrix:
+    - `S` (single-precision real: `float`)
+    - `D` (double-precision real: `double`)
```
tfalders, line 36: It might also be worth mentioning the complex-valued versions of the functions: Cheevdx and Zheevdx

Libraries/hipSOLVER/syevdx/README.md
```diff
@@ -0,0 +1,132 @@
...
+    - `int il`: Smallest index of the eigenvalues to be returned if `range` = `HIPSOLVER_EIG_RANGE_I`.
+    - `int iu`: Largest index of the eigenvalues to be returned if `range` = `HIPSOLVER_EIG_RANGE_I`.
+    - `int *nev`: Number of eigenvalues returned.
+    - `double *W`: Pointer to array $W$ in device memory, where the resulting eigenvalues are written.
+    - `double *work`: Pointer to working space in device memory.
+    - `int lwork`: Size of the working space.
+    - `int *devInfo`: Convergence result of the function in device memory.
+        - If 0, the algorithm converged.
+        - If greater than 0 (`devInfo = i` for $0 < i \leq n$), then `devInfo` eigenvectors did not converge.
+        - If lesser than 0 (`devInfo = -i` for $0 < i \leq n$), then the the $i^{th}$ parameter is wrong (not counting the handle).
```
tfalders: This is only the case for the cuSOLVER backend. With the rocSOLVER backend, devInfo >= 0.

Libraries/hipSOLVER/syevdx/main.cpp
```diff
@@ -0,0 +1,208 @@
...
+        errors++;
+        goto cleanup;
+    }
+    else if(syevdx_info > 0)
+    {
+        std::cout << "Computing eigenvalues did not converge.\n" << std::endl;
+        errors++;
+        goto cleanup;
+    }
+    {
```
tfalders: Should there be an 'else` between these brackets?

Libraries/hipSOLVER/syevj/README.md
```diff
@@ -0,0 +1,92 @@
...
+  - `HIPSOLVER_EIG_MODE_VECTOR`: compute the eigenvalues and eigenvectors.
+  - `HIPSOLVER_EIG_MODE_NOVECTOR`: only compute the eigenvalues.
+- `hipsolverFillMode_t`: specifies which part of $A$ to use.
+  - `HIPSOLVER_FILL_MODE_LOWER`: data is stored in the lower triangle of $A$ .
+  - `HIPSOLVER_FILL_MODE_UPPER`: data is stored in the upper triangle of $A$ .
+- `hipsolverCreateSyevjInfo`: initializes a structure for the parameters and results for calling `syevj`.
+- `hipsolverDestroySyevjInfo`: destroys the structure for the parameters and results for calling `syevj`.
+- `hipsolverXsyevjSetMaxSweeps`: configures the max amounts of sweeps
+- `hipsolverXsyevjSetTolerance`: configures  the tolerance of `syevj`.
+- `hipsolverXsyevjSetSortEig` : configures wether to sort the results or not
```
tfalders: Typo on whether

Libraries/hipSOLVER/syevj/main.cpp
```diff
@@ -0,0 +1,190 @@
...
+    std::default_random_engine             generator;
+    std::uniform_real_distribution<double> distribution(0., 2.);
+    auto                                   random_number = std::bind(distribution, generator);
+
+    for(int i = 0; i < n; i++)
+    {
+        A[(n + 1) * i] = random_number();
+        for(int j = 0; j < i; j++)
+        {
+            A[i * n + j] = A[j * n + i] = random_number();
```
tfalders: I would suggest using lda instead of n in case someone wants to try changing lda

Libraries/hipSOLVER/syevj/main.cpp
```diff
@@ -0,0 +1,190 @@
...
+    std::cout << "A = " << format_range(A.begin(), A.end()) << std::endl;
+    std::cout << "W = " << format_range(W.begin(), W.end()) << std::endl;
+    std::cout << "V = " << format_range(V.begin(), V.end()) << std::endl;
+
+    int    sweeps   = 0;
+    double residual = 0;
+    HIPSOLVER_CHECK(hipsolverXsyevjGetSweeps(hipsolver_handle, params, &sweeps));
+    HIPSOLVER_CHECK(hipsolverXsyevjGetResidual(hipsolver_handle, params, &residual));
+
+    std::cout << "\nWhich was computed in " << sweeps << " sweeps, with a residue of " << residual
```
tfalders: residue -> residual

Libraries/hipSOLVER/syevj/main.cpp
```diff
@@ -0,0 +1,190 @@
...
+              << std::endl;
+
+    // 11. Validate that 'AV == VD' and that 'AV - VD == 0'.
+    std::cout << "\nLet D be the diagonal constructed from W.\n"
+              << "The right multiplication of A * V should result in V * D [AV == VD]:"
+              << std::endl;
+
+    // Right multiplication of the input matrix with the eigenvectors.
+    std::vector<double> AV(n * n);
+    multiply_matrices(1.0, 0.0, n, n, n, A.data(), n, 1, V.data(), 1, n, AV.data(), n);
```
tfalders: If lda is used up above, it should be used here too. It might be worth considering giving each matrix its own leading dimension (e.g. ldv), but I leave that up to you.

Libraries/hipSOLVER/syevj/main.cpp
```diff
@@ -0,0 +1,190 @@
...
+    // Construct the diagonal D from eigenvalues W.
+    std::vector<double> D(n * n);
+    for(int i = 0; i < n; i++)
+    {
+        D[(n + 1) * i] = W[i];
+    }
+
+    // Scale eigenvectors V with W by multiplying V with D.
+    std::vector<double> VD(n * n);
+    multiply_matrices(1.0, 0.0, n, n, n, V.data(), 1, n, D.data(), n, 1, VD.data(), n);
```
tfalders: If lda is used up above, it will need to be used for V here too

Libraries/hipSOLVER/syevj_batched/README.md
```diff
@@ -0,0 +1,133 @@
...
+  - `int n`: number of rows and columns of each $A_i$.
+  - `double* A`: pointer to the input batch $A$ on device memory.
+  - `int lda`: leading dimension of the matrices $A_i$.
+  - `double* W`: pointer to array $W$ in device memory, where the resulting eigenvalues are written.
+  - `double* work`: pointer to working space in device memory.
+  - `int lwork`: size of the working space.
+  - `int* devInfo`: pointer to where the convergence result of the function is written to in device memory.
+        - If 0, the algorithm converged.
+        - If greater than 0 (`devInfo = i` for $1 \leq i \leq n$), then `devInfo` eigenvectors did not converge.
+        - If lesser than 0 (`devInfo = -i` for $1 \leq i \leq n$), then the the $i^{th}$ parameter is wrong (not counting the handle).
```
tfalders: Only for the cuSOLVER backend. With the rocSOLVER backend, devInfo >= 0.

Libraries/hipSOLVER/syevj_batched/main.cpp
```diff
@@ -0,0 +1,248 @@
...
+        return 0;
+    }
+    const int lda         = n;
+    const int size_matrix = n * lda;
+
+    // Get the batch size.
+    const int batch_count = parser.get<int>("c");
+    if(batch_count <= 0)
+    {
+        std::cout << "Batch size should be at least 1" << std::endl;
```
tfalders, line 60: if(batch_count < 1)?

Libraries/hipSOLVER/syevj_batched/main.cpp
```diff
@@ -0,0 +1,248 @@
...
+    auto random_number = [&]() { return distribution(generator); };
+
+    for(int k = 0; k < batch_count * size_matrix; k += size_matrix)
+    {
+        for(int i = 0; i < n; ++i)
+        {
+            A[k + (n + 1) * i] = random_number();
+            for(int j = 0; j < i; ++j)
+            {
+                A[k + i * n + j] = A[k + j * n + i] = random_number();
```
tfalders: Again, I recommend using lda instead of n

Libraries/hipSOLVER/sygvj/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+where
+- $A,B\in\mathbb{R}^{n\times n}$ are symmetric matrices,
+- $\lambda_i$ for $i\in\{1, \dots n\}$ are eigenvalues,
+- $v_i\in\mathbb{R}^n$ are eigenvectors corresponding to the $i$-th eigenvalue (they can be normalized to unit length).
+
+This choice corresponds to `HIPSOLVER_EIG_TYPE_1` parameter value of the solver function `hipsolverDsygvj`. Two other possibilities include $ABv_i = \lambda_i v_i$ for `HIPSOLVER_EIG_TYPE_2` and $BAv_i = \lambda_i v_i$ for `HIPSOLVER_EIG_TYPE_3`.
+
+### Application flow
+1. Instantiate two vectors containing $A$'s and $B$'s 9 elements.
```
tfalders: Probably good to mention that the value for n is 3

Libraries/hipSOLVER/sygvj/README.md
```diff
@@ -0,0 +1,107 @@
...
+10. Copy residual and executed sweeps number to the host. Print their values.
+11. Free all allocated resources.
+
+
+## Key APIs and Concepts
+### hipSOLVER
+- hipSOLVER is initialized by calling `hipsolverCreate(hipsolverHandle_t*)` and it is terminated by calling `hipsolverDestroy(hipsolverHandle_t)`.
+- `hipsolver[SD]sygvj` computes the generalized eigenvalues of an $n \times n$ matrix pair $A$ and $B$. The correct function signature should be chosen based on the datatype of the input matrix::
+    - `S` (single-precision real: `float`)
+    - `D` (double-precision real: `double`)
```
tfalders, line 40: I'd suggest mentioning the complex versions (Chegvj and Zhegvj) as well

Libraries/hipSOLVER/sygvj/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+    In this example, a double-precision real input matrix pair is used, in which case the function accepts the following parameters:
+    - `hipsolverHandle_t handle`
+    - `hipsolverEigType_t itype`: Specifies the type of eigensystem problem, see [above](#description).
+    - `hipsolverEigMode_t jobz`: Specifies whether the eigenvectors should also be calculated besides the eigenvalues. The following values are accepted:
+        - `HIPSOLVER_EIG_MODE_NOVECTOR`: Calculate the eigenvalues only.
+        - `HIPSOLVER_EIG_MODE_VECTOR`: Calculate both the eigenvalues and the eigenvectors. The eigenvectors are calculated by a divide and conquer algorithm and are written to the memory location specified by `*A`.
```
tfalders: I don't think the eigenvectors are calculated by divide-and-conquer (at least not with the rocSOLVER backend). IIRC, they are computed using the Jacobi method by applying Jacobi rotations to an initial identity matrix.

Libraries/hipSOLVER/sygvj/README.md
```diff
@@ -0,0 +1,107 @@
...
+- `hipsolverCreate`
+- `hipsolverDestroy`
+- `hipsolverDnCreateSyevjInfo`
+- `hipsolverDnDestroySyevjInfo`
+- `hipsolverDnXsyevjSetTolerance`
+- `hipsolverDnXsyevjSetMaxSweeps`
+- `hipsolverDnXsyevjGetResidual`
+- `hipsolverDnXsyevjGetSweeps`
+- `hipsolverDsygvj_bufferSize`
+- `hipsolverDsygvj`
```
tfalders: It looks like you're mixing calls to the compatibility API with calls to the regular API, which is perfectly valid but it would be nice to be consistent.

Libraries/hipSOLVER/sygvj/main.cpp
```diff
@@ -0,0 +1,202 @@
...
+        std::vector<double> V(size_A, 0);
+        HIP_CHECK(hipMemcpy(V.data(), d_A, sizeof(double) * size_A, hipMemcpyDeviceToHost));
+
+        std::cout << "Eigenvectors:" << std::endl;
+        for(int i = 0; i < n; ++i)
+        {
+            std::cout << "{ ";
+            for(int j = 0; j < n; ++j)
+            {
+                std::cout << V[i * n + j] << (j < n - 1 ? ", " : " }\n");
```
tfalders: Here's another place to use lda instead of n

Libraries/hipSOLVER/sygvj/main.cpp
```diff
@@ -0,0 +1,202 @@
...
+    }
+    else
+    {
+        if(syevj_info < 0)
+        {
+            std::cout << "Parameter number " << -syevj_info << " is wrong.";
+        }
+        else if(syevj_info <= n)
+        {
+            std::cout << "Leading minor of order " << -syevj_info
```
tfalders: I don't think you need to negate syevj_info here

Libraries/hipSOLVER/sygvj/main.cpp
```diff
@@ -0,0 +1,202 @@
...
+                      << " of B is not positive definite.";
+        }
+        else if(syevj_info == n + 1)
+        {
+            std::cout << "Sygvj doen not converge, error " << syevj_info << ".";
```
tfalders: Typo on does

tfalders (commented):
Looks good! Just a few comments to be addressed before I give my approval.

Libraries/hipSOLVER/syevdx/README.md
Beanavil: Done!

Libraries/hipSOLVER/syevdx/README.md
Beanavil: True, added this info to the README.

Libraries/hipSOLVER/syevdx/main.cpp
Beanavil: Originally there was an `else`, but then the following suggestion was made:
>Something like
>```c++
>if (/* some condition */) {
>  // clean and return
>}
>
>// continue program
>// cleanup
>```
>or
>```c++
>if (/* some condition */) {
>  goto cleanup_and_return;
>}
>
>// continue program
>
>cleanup_and_return:
>// cleanup
>```
>is generally preferred over:
>```c++
>if (/* inverse of some condition*/) {
>  // continue program
>}
>// cleanup
>```
>as it reduces nesting/indentation.

But after choosing Option 2 (goto approach) removing the braces resulted in a `cannot jump from this goto statement to its label` error, because then the jump bypassed the `eps`, `h_one`. `h_minus_one` and `h_zero` variables initialization. So the braces were left to prevent this error from appearing.

Maybe you have some other opinion about this design?

Libraries/hipSOLVER/syevj/README.md
Beanavil: Nice catch, fixed (also in syevj_batched).

Libraries/hipSOLVER/syevj/main.cpp
Beanavil: Fixed.

Libraries/hipSOLVER/syevj/main.cpp
Beanavil: Nice catch, fixed.

Libraries/hipSOLVER/syevj/main.cpp
Beanavil: Added the use of `lda`. Also for the size of AV here and in the `syev` example.

Libraries/hipSOLVER/syevj/main.cpp
Beanavil: Added the use of `lda` here as well.  Also for the size of VD here and in the `syev` example.

Libraries/hipSOLVER/syevj_batched/README.md
Beanavil: Added this info to the README.

Libraries/hipSOLVER/syevj_batched/main.cpp
Beanavil: It should be the same in this case, given that it's an integer. Or do you prefer it that way for some other reason?

Libraries/hipSOLVER/syevj_batched/main.cpp
Beanavil: Fixed.

Libraries/hipSOLVER/sygvj/README.md
Beanavil: Changed!

Libraries/hipSOLVER/sygvj/README.md
Beanavil: Done!

Libraries/hipSOLVER/sygvj/README.md
Beanavil: True, must have been some copy-paste error. Fixed!

Libraries/hipSOLVER/sygvj/README.md
Beanavil: Oopsie, we shouldn't be mixing these. Should be fixed now.

Libraries/hipSOLVER/sygvj/main.cpp
Beanavil: Fixed!

Libraries/hipSOLVER/sygvj/main.cpp
Beanavil: True, fixed.

Libraries/hipSOLVER/sygvj/main.cpp
Beanavil: Nice catch, fixed!

Libraries/hipSOLVER/syevj_batched/main.cpp
tfalders: The if statement and the error message don't match. I don't mind leaving it as `batch_count <= 0`, as long as the error message specifies the correct condition.

Libraries/hipSOLVER/syevdx/main.cpp
tfalders: Since the braces have to be kept anyway, I don't think you're saving anything on indentation. Using else instead of goto would also make the code more consistent with the syevj_batched example.

Libraries/hipSOLVER/syevj_batched/main.cpp
Beanavil: Do you mean something like "Batch size should be greater than 0"?

Libraries/hipSOLVER/syevdx/main.cpp
Beanavil: Okey, should be changed now. BTW, `sygvd` was also changed so it's consistent with that approach.

Libraries/hipSOLVER/syevj_batched/main.cpp
tfalders: Oh! Silly me, I misinterpreted what would happen when `batch_count = 0`. My mistake. No change needed.

tfalders (approved):
Nice work! Approved!

Naraenda (approved):

PR #48, opened 2023-09-15
fix: all broken readme links pointing to examples

### LINKED ISSUE #15: Links broken in the Readme
Please ensure the links work in the README.md file.

Also, please rearrange the headers so we start with a list of examples as the first heading. The following heading should list the other folders, common, docker, etc. Showcase the examples first.

### LINKED ISSUE #44: How to get access to example?
Hi,

Maybe a stupid question, it seems I don't have access to https://github.com/Applications/floyd_warshall/ (which presents here https://github.com/amd/rocm-examples/blob/develop/README.md)with 404 error. 

Is it a private repo? How do I join the deverloper orgnization in order to get access to examples shown here https://github.com/amd/rocm-examples/blob/develop/README.md

LucasSnatiago:
The readme file is pointing to the root of the Github website. I changed the links to be pointing to the repository folders. It should solve the Page Not Found issue when clicking on the readme links. That's not the fanciest solution I know, but it's better than just not working at all.

It only works on the main branch, because of the [markdown limitations](https://github.com/github/markup/issues/913).

fix: #15 
fix: #44

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,102 +1,102 @@
 # ROCm Examples
 This project is currently unsupported and in an early testing stage. Feedback on the contents of this repository is appreciated.
 ## Repository Contents
-- [Applications](/Applications/) groups a number of examples ... .
-    - [floyd_warshall](/Applications/floyd_warshall/): Showcases a GPU implementation of the Floyd-Warshall algorithm for finding shortest paths in certain types of graphs.
-- [Common](/Common/) contains common utility functionality shared between the examples.
-- [HIP-Basic](/HIP-Basic/) hosts self-contained recipes showcasing HIP runtime functionality.
-    - [assembly_to_executable](/HIP-Basic/assembly_to_executable): Program and accompanying build systems that show how to manually compile and link a HIP application from host and device code.
-    - [bandwidth](/HIP-Basic/bandwidth): Program that measures memory bandwidth from host to device, device to host, and device to device.
-    - [bit_extract](/HIP-Basic/bit_extract): Program that showcases how to use HIP built-in bit extract.
-    - [device_globals](/HIP-Basic/device_globals): Show cases how to set global variables on the device from the host.
-    - [device_query](/HIP-Basic/device_query): Program that showcases how properties from the device may be queried.
-    - [dynamic_shared](/HIP-Basic/dynamic_shared): Program that showcases how to use dynamic shared memory with the help of a simple matrix transpose kernel.
-    - [events](/HIP-Basic/events/): Measuring execution time and synchronizing with HIP events.
-    - [gpu_arch](/HIP-Basic/gpu_arch/): Program that showcases how to implement GPU architecture-specific code.
-    - [hello_world](/HIP-Basic/hello_world): Simple program that showcases launching kernels and printing from the device.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,102 +1,102 @@
...
-        - [norm](/Libraries/rocThrust/norm/): An example that computes the Euclidean norm of a `thrust::device_vector`.
-        - [reduce_sum](/Libraries/rocThrust/reduce_sum/): An example that computes the sum of a `thrust::device_vector` integer vector using the `thrust::reduce()` generalized summation and the `thrust::plus` operator.
-        - [remove_points](/Libraries/rocThrust/remove_points/): Simple program that demonstrates the usage of the `thrust` random number generation, host vector, generation, tuple, zip iterator, and conditional removal templates. It generates a number of random points in a unit square and then removes all of them outside the unit circle.
-        - [saxpy](/Libraries/rocThrust/saxpy/): Simple program that implements the SAXPY operation (`y[i] = a * x[i] + y[i]`) using rocThrust and showcases the usage of the vector and functor templates and of `thrust::fill` and `thrust::transform` operations.
-        - [vectors](/Libraries/rocThrust/vectors/): Simple program that showcases the `host_vector` and the `device_vector` of rocThrust.
+- [Applications](/amd/rocm-examples/tree/develop/Applications/) groups a number of examples ... .
```
chicheng: how about relative path? 

-  `[Applications](./Applications/)` (add leading `.`) [Applications](./Applications/) , or
-  `[Applications](Applications/)` (remove leading `/`) [Applications](Applications/)

LucasSnatiago: I tried but it seems to not work on forks it keeps pointing to the wrong path, so I couldn't test it. Hardcoding the link was the only way to be 100% sure that the examples would work, but if you think that's better I can change my PR.

Naraenda:
The current method seems to only work on GitLab. It seems the correct way to have the Markdown work properly on GitHub would be to include the `README.md` in the URL. The change would be 
```diff
- - [Applications](/Applications/) groups a number of examples ... .
+ - [Applications](Applications/README.md) groups a number of examples ... .
```
The latter would render as 
- [Applications](Applications/README.md) groups a number of examples ... .

--- 

If you could update the PR with that, that would be nice. Thanks!

LucasSnatiago:
I updated the PR yesterday. Should be working now!

Maetveis:
> I updated the PR yesterday. Should be working now!

Still doesn't work if you try it from the project main page instead of clicking on the README and then using the links.

I've messed about with using various types of redirects and I think GitHub is just broken, and we're hitting the same issue as https://github.com/orgs/community/discussions/67750.

LucasSnatiago:
Nice research!

I made a lot of tests and the only one that worked for me was to make absolute paths and not relative paths. I thought it was just me that did something wrong, but apparently it's not just me. First I made a PR using only absolute paths. I think it is better to have something working being hardcoded than not have anything working at all. Even though I still agree with both @chicheng and @Naraenda that the right way to do it is using relative paths.

I think we can leave this PR as it is, so the next people looking for the reason of the README links not working can understand reading this discussion.

saadrahim:
@Naraenda irrespective of the GitHub bug, can you have a plan to fix this? Even if you have absolute links, we cannot have this project be broken any longer.

saadrahim:
Closed with #51

Issue #89, opened 2024-01-23
Can't build with ROCm/HIP 6.0.0


PPPW:
This repo worked fine for me with ROCm/HIP 5.6.0, however when I'm trying to build it (just run `make`) with ROCm/HIP 6.0.0, I got:
```
make[2]: Entering directory '/opt/rocm-examples/HIP-Basic/opengl_interop'
/opt/rocm/bin/hipcc -std=c++17 -Wall -Wextra -I ../../Common -I ../../External   -o hip_opengl_interop main.hip ../../External/glad/glad.cpp -lglfw -ldl
main.hip:104:65: error: use of undeclared identifier 'hipGLDeviceListAll'
    HIP_CHECK(hipGLGetDevices(&gl_device_count, &hip_device, 1, hipGLDeviceListAll));
                                                                ^
main.hip:428:19: error: use of undeclared identifier 'hipGraphicsGLRegisterBuffer'
        HIP_CHECK(hipGraphicsGLRegisterBuffer(
                  ^
2 errors generated when compiling for gfx906.
make[2]: *** [Makefile:58: hip_opengl_interop] Error 1
make[2]: Leaving directory '/opt/rocm-examples/HIP-Basic/opengl_interop'
make[1]: *** [Makefile:64: opengl_interop] Error 2
make[1]: Leaving directory '/opt/rocm-examples/HIP-Basic'
make: *** [Makefile:34: HIP-Basic] Error 2
```

I searched "hipGraphicsGLRegisterBuffer" and "hipGLDeviceListAll", [they exist](https://docs.amd.com/projects/HIP/en/docs-5.7.0/.doxygen/docBin/html/group___g_l.html) in previous HIP versions however not in [latest HIP version](https://docs.amd.com/projects/HIP/en/docs-6.0.0/doxygen/html/group___g_l.html). Do we have plans to make this repo work with ROCm 6.0.0? Thanks

Naraenda:
This is probably fixed with https://github.com/ROCm/rocm-examples/pull/99

PPPW:
> This is probably fixed with #99

Will verify once that PR is merged.

PR #90, opened 2024-01-25
 rocSPARSE examples (COO, ELL and GEBSR formats)

Beanavil:
This pull request contains the third batch of rocSPARSE examples. This group of examples corresponds to the ones that showcase COO-, ELL-. and GEBSR-formats-related functionality, as well as some small fixes to previously added samples.

Examples added:
- level 2
  - `coomv`
  - `ellmv`
  - `gebsrmv`
- level 3
  - `gebsrmm`


Libraries/rocSPARSE/level_2/bsrmv/README.md
```diff
@@ -13,8 +13,8 @@ where
 - $A'$ is a sparse matrix in BSR format with `rocsparse_operation` and described below.
 
 ## Application flow
-1. Setup a sparse matrix in BSR format. Allocate an x and a y vector and set up $\alpha$ and $\beta$ scalars.
-2. Setup a handle, a matrix descriptor and a matrix info.
```
emankov, line 17: It is better to separate feature and formatting commits. There is no need to make already big PRs even bigger.

Libraries/rocSPARSE/level_2/bsrmv/main.cpp
```diff
@@ -145,7 +146,7 @@ int main()
                                         &beta,
                                         d_y));
 
-    // 5. Copy y to host from device
```
emankov: Here and below, I'd suggest creating a separate PR dedicated to `bsr(x)mv` fixes.

Libraries/rocSPARSE/level_2/bsrxmv/bsrxmv_vs2017.vcxproj
```diff
@@ -15,7 +15,7 @@
     <VCProjectVersion>15.0</VCProjectVersion>
     <ProjectGuid>{CFA1826D-C202-4F97-92C9-DB019675DCCB}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
-    <RootNamespace>bsrmv_vs2017</RootNamespace>
```
emankov, line 18: Thanks for catching this. Nonetheless - to a separate `bsr(x)mv` PR.

Libraries/rocSPARSE/level_2/coomv/main.cpp
```diff
@@ -0,0 +1,140 @@
...
+    // Matrix descriptor
+    rocsparse_mat_descr descr;
+    ROCSPARSE_CHECK(rocsparse_create_mat_descr(&descr));
+
+    // 3. Offload data to device
+    rocsparse_int* d_coo_row_ind;
+    rocsparse_int* d_coo_col_ind;
+    double*        d_coo_val;
+    double*        d_x;
+    double*        d_y;
```
emankov: should be initialized

Libraries/rocSPARSE/level_2/ellmv/main.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+    // Matrix descriptor
+    rocsparse_mat_descr descr;
+    ROCSPARSE_CHECK(rocsparse_create_mat_descr(&descr));
+
+    // 3. Offload data to device
+    rocsparse_int* d_ell_col_ind;
+    double*        d_ell_val;
+    double*        d_x;
+    double*        d_y;
```
emankov: should be initialized

Libraries/rocSPARSE/level_2/gebsrmv/main.cpp
```diff
@@ -0,0 +1,195 @@
...
+    constexpr std::array<double, n_padded> h_x{1.0, 2.0, 3.0, 4.0, 0.0, 0.0};
+    std::array<double, m_padded>           h_y{4.0, 5.0, 6.0, 7.0};
+    constexpr std::array<double, m_padded> expected_y{0.0, 5.0, 10.0, 13.0};
+
+    // 2. Allocate device memory and offload input data to device.
+    rocsparse_int* d_bsr_row_ptr;
+    rocsparse_int* d_bsr_col_ind;
+    double*        d_bsr_val;
+    double*        d_x;
+    double*        d_y;
```
emankov: should be initialized

Libraries/rocSPARSE/level_2/gebsrmv/main.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+    // 8. Print results to standard output.
+    std::cout << "Solution successfully computed: ";
+
+    std::cout << "y = " << format_range(std::begin(h_y), std::end(h_y)) << std::endl;
+
+    // Compare solution with the expected result.
+    int          errors{};
+    const double eps = 1.0e5 * std::numeric_limits<double>::epsilon();
+    for(size_t i = 0; i < h_y.size(); i++)
```
emankov: ++i

Libraries/rocSPARSE/level_3/gebsrmm/main.cpp
```diff
@@ -0,0 +1,209 @@
...
+    // Matrix descriptor
+    rocsparse_mat_descr descr;
+    ROCSPARSE_CHECK(rocsparse_create_mat_descr(&descr));
+
+    // 3. Offload data to device
+    rocsparse_int* d_bsr_row_ptr;
+    rocsparse_int* d_bsr_col_ind;
+    double*        d_bsr_val;
+    double*        d_B;
+    double*        d_C;
```
emankov: should be initialized

Libraries/rocSPARSE/preconditioner/bsric0/main.cpp
```diff
@@ -184,7 +188,7 @@ int main()
                                       bsr_dim,
                                       info,
                                       solve_policy,
-                                      temp_buffer))
```
emankov, line 187: I wasn't even built correctly, right?

Libraries/rocSPARSE/level_3/gebsrmm/.gitignore
```diff
@@ -0,0 +1 @@
+rocsparse_gebsrmm
```
emankov, line 1: Why should it be ignored?

emankov (changes requested):
Thanks for your multiple catches, which are better to move to a separate PR. 
Btw, did you build and run all those newly added projects?

Libraries/rocSPARSE/level_2/bsrmv/README.md
Beanavil: So then I can drop the [Fix and clarify rocsparse examples device synchronization](https://github.com/amd/rocm-examples/pull/90/commits/d43ffd6bbdf88e25119388753b9bf5d82f7dac4d) commit, I think it's the one introducing this many formatting/comment fixes

Libraries/rocSPARSE/level_3/gebsrmm/.gitignore
Beanavil: This is only the executable generated after compiling, we have a .gitignore inside each example's folder so that we don't actually push the executables to the repo

Libraries/rocSPARSE/preconditioner/bsric0/main.cpp
Beanavil: The commit is precisely adding the semicolon at the end of the call cause it was causing compiling errors:
image
 I will keep this change to ensure that they build

Libraries/rocSPARSE/level_3/gebsrmm/main.cpp
Beanavil: Done! Empty initialized

Libraries/rocSPARSE/level_2/gebsrmv/main.cpp
Beanavil: Fixed!

Libraries/rocSPARSE/level_2/gebsrmv/main.cpp
Beanavil: Done! Empty initialized

Libraries/rocSPARSE/level_2/ellmv/main.cpp
Beanavil: Done! Empty initialized

Libraries/rocSPARSE/level_2/coomv/main.cpp
Beanavil: Done! Empty initialized

Libraries/rocSPARSE/level_2/bsrxmv/bsrxmv_vs2017.vcxproj
Beanavil: Done! I will open a cleanup PR with several commits that include comments, documentation and some other fixes to the sparse examples

Libraries/rocSPARSE/level_2/bsrmv/main.cpp
Beanavil: Done!

Beanavil:
Addresing this question
> Btw, did you build and run all those newly added projects?

The examples build for ROCm 5.7 and for ROCm 6.0, but for the latter we still need to open a PR upstreaming some fixes for breaking stuff after upgrading to ROCm 6.0 (like in [this issue](https://github.com/amd/rocm-examples/issues/89)).

Also, in Windows, we are seeing some error messages concerning visual studio `.vcxproj` files for ROCm 5.7 with our current SDK version (23.30_ROCM5.7_V46), but we need to update it to the latest version we received so tomorrow I can give more insights about this.

Update @emankov: after installing the latest version of the SDK that we have (23.30_ROCM5.7_V66) we got a build error that came from the `HIP_Basic/module_api` example, as the VS variable `OutDir` is not defined anymore. This can be solved by adding some code to the `.vcxproj` files of this example, but this was not necessary before so perhaps this should be looked into from AMD's side.
For the tests run, some tests are failing for ROCm and all tests fail for CUDA with this error: `-1073741515` when run with Visual Studio.

YvanMokwinski:
Would it be suitable to include anyone from the rocSPARSE team as reviewer(s)?

Libraries/rocSPARSE/level_2/gebsrmv/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+    HIP_CHECK(hipFree(d_y));
+
+    // 8. Print results to standard output.
+    std::cout << "Solution successfully computed: ";
+
+    std::cout << "y = " << format_range(std::begin(h_y), std::end(h_y)) << std::endl;
+
+    // Compare solution with the expected result.
+    int          errors{};
+    const double eps = 1.0e5 * std::numeric_limits<double>::epsilon();
```
YvanMokwinski, line 186: 1e5 factor looks a bit random.

Libraries/rocSPARSE/level_2/gebsrmv/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+    for(size_t i = 0; i < h_y.size(); ++i)
```
YvanMokwinski, line 187: Examples are not consistent with providing an error check with the expected result. Ellmv is different from this, for instance.

YvanMokwinski: Moreover, results should be exact here since inputs are integer values. The eps sounds random, and the relative error should be computed rather than absolute.

YvanMokwinski (approved):
I would like the examples to be consistent with checking the expected results.

Libraries/rocSPARSE/level_2/gebsrmv/main.cpp
Beanavil: For the first remark, the error-checking has been homogenized across examples in the last commit 👍🏼 . 
For the second one, while it's true that in this case the values of the matrix happen to be integers, the type used for the matrix' values is `double`, so I think we should still showcase the proper error checking for double precision floats. Perhaps the input matrix/scalars could be changed so that floating point values are actually used.

YvanMokwinski: According to the inputs, you can set the tolerance to zero here; but that's a detail.
It is good that error-checking has been homogenized across examples.

emankov (approved):
LGTM

PR #121, opened 2024-06-03
Add cmake-format to CI

dgaliffiAMD:


Naraenda:
`cmake-format` seems to be no longer maintained: [last update 3 years ago](https://github.com/cheshirekow/cmake_format). Can something like [`gersemi`](https://github.com/BlankSpruce/gersemi) be a better alternative?

dgaliffiAMD:
> `cmake-format` seems to be no longer maintained: [last update 3 years ago](https://github.com/cheshirekow/cmake_format). Can something like [`gersemi`](https://github.com/BlankSpruce/gersemi) be a better alternative?

Thanks! I'm not familiar with `gersemi`, outside of Norse mythology. The usage looks similar to `cmake-format`, so it should be an easy patch.  I'll look into it. :)

Naraenda (approved):
LGTM, thanks!

Beanavil (approved):

dgaliffiAMD:
Thanks for the tip, @Naraenda . Gersemi was nice to work with. Similar to cmake-format but with some nice quality-of-life improvements, like accepting the directory as a source.

PR #122, opened 2024-06-04
Fixed path in error message.

dgaliffiAMD:
'scripts/copyright-date/check-copyright.sh' is no longer the correct path to the file. Using "${BASH_SOURCE[0]}" instead.


Beanavil (approved):
LGTM

PR #139, opened 2024-07-05
Add NO_DEFAULT_PATH for llvm-mc, llvm-dis, clang-offload-bundler

dgaliffiAMD:
Found in QA image, where LLVM is installed, but rocm-llvm-dev is not. `/usr/bin/llvm-dis` and `/opt/rocm/llvm/bin/llvm-mc` were being used to build the `llvm_ir_to_executable` example. This version mismatch results in a build failure.

Adding `NO_DEFAULT_PATH` to only search for these programs in the specified paths. ${ROCM_PATH}, for example.


Snektron (changes requested):
This should probably also be fixed in HIP-Basic/assembly_to_executable/CMakeLists.txt and HIP-Basic/llvm_ir_to_executable/CMakeLists.txt

Snektron (approved):

PR #140, opened 2024-07-12
Add source format linting to the GitHub workflows

Mustaballer:
Work on standardizing `git-clang-format` for both pre-commit hooks and CI checks to avoid discrepancies.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -53,7 +53,9 @@ jobs:
           sudo echo "deb-src http://apt.llvm.org/jammy/ llvm-toolchain-jammy-18 main" | sudo tee -a /etc/apt/sources.list.d/llvm-18.list
           sudo apt-get update -qq
           sudo apt-get install -y clang-format-18
-          sudo apt-get install -y clang-tools-18
+          sudo apt-get install -y clang-tools-18  # Install git-clang-format
+      - name: Fetch all branches
+        run: git fetch --all
       - name: Run clang-format
         env:
           GIT_CLANG_FORMAT: /usr/lib/llvm-18/bin/git-clang-format
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -61,4 +61,4 @@ jobs:
           GIT_CLANG_FORMAT: /usr/lib/llvm-18/bin/git-clang-format
         run: |
           git config --global --add safe.directory $GITHUB_WORKSPACE
-          Scripts/CodeFormat/check_format.sh origin/`if [[ "${{ github.event_name }}" == "push" ]]; then echo "develop"; else echo "${{ github.base_ref }}"; fi` --binary ${{env.CLANG_FORMAT}}
+          Scripts/CodeFormat/check_format.sh --all-files origin/`if [[ "${{ github.event_name }}" == "push" ]]; then echo "develop"; else echo "${{ github.base_ref }}"; fi` --binary ${{env.CLANG_FORMAT}}
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -35,3 +35,18 @@ jobs:
             echo "For example: \`gersemi -i .\`"
             exit 1
           fi
+
+  source-formatting:
+    name: Source Code Formatting
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y clang-format
+      - name: Run clang-format
+        run: |
+          FILES=$(find . -type f ! -path '**/build/*' | egrep '\.(cc|cp|cpp|c++|cxx|cu|cuh|hh|hpp|hxx|hip|vert|frag)$')
+          clang-format -style=file -i $FILES
+          git diff --exit-code
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -53,8 +53,10 @@ jobs:
           sudo echo "deb-src http://apt.llvm.org/jammy/ llvm-toolchain-jammy-18 main" | sudo tee -a /etc/apt/sources.list.d/llvm-18.list
           sudo apt-get update -qq
           sudo apt-get install -y clang-format-18
+          sudo apt-get install -y clang-tools-18
       - name: Run clang-format
+        env:
+          GIT_CLANG_FORMAT: /usr/lib/llvm-18/bin/git-clang-format
         run: |
           git config --global --add safe.directory $GITHUB_WORKSPACE
           Scripts/CodeFormat/check_format.sh origin/`if [[ "${{ github.event_name }}" == "push" ]]; then echo "develop"; else echo "${{ github.base_ref }}"; fi` --binary ${{env.CLANG_FORMAT}}
-          
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -45,8 +45,16 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y clang-format
-      - name: Run clang-format
+      - name: Make scripts executable
+        run: chmod +x Scripts/CodeFormat/check_format.sh Scripts/CopyrightDate/check_copyright.sh
+      - name: Run code format check
+        run: ./Scripts/CodeFormat/check_format.sh HEAD
+      - name: Run copyright check
         run: |
-          FILES=$(find . -type f ! -path '**/build/*' | egrep '\.(cc|cp|cpp|c++|cxx|cu|cuh|hh|hpp|hxx|hip|vert|frag)$')
-          clang-format -style=file -i $FILES
-          git diff --exit-code
+          opts="-qc"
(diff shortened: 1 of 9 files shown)


Mustaballer:
@dgaliffiAMD Ready for review!

example.cpp
dgaliffiAMD, line 1: Don't forget to remove your `example.cpp` and squash merge, so it does not become part of the repository's history.

Scripts/CodeFormat/check_format.sh
dgaliffiAMD, line 1: It looks like you left some debug logs in the script file.

dgaliffiAMD (changes requested):
Thanks @Mustaballer. I'll test it out on my system, but I've made some notes for now.

.github/workflows/linting.yml
```diff
@@ -35,3 +35,18 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y clang-format-14
+      - name: Run clang-format
+        run: |
+          FILES=$(find . -type f ! -path '**/build/*' | egrep '\.(cc|cp|cpp|c++|cxx|cu|cuh|hh|hpp|hxx|hip|vert|frag)$')
+          clang-format -style=file -i $FILES
+          git diff --exit-code
```
Beanavil: It's probably better to use the latest version. I'd suggest doing something like:
```diff
+ env:
+   CLANG_FORMAT: /usr/lib/llvm-18/bin/clang-format

jobs:
  ...

  source-formatting:
    name: Source Code Formatting
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Install dependencies
        run: |
-           sudo apt-get update
-           sudo apt-get install -y clang-format-14
+           sudo apt-get update -qq
+           apt-get install -y git wget
+           wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc
+           sudo echo "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-18 main" | sudo tee -a /etc/apt/sources.list.d/llvm-18.list
+           sudo echo "deb-src http://apt.llvm.org/jammy/ llvm-toolchain-jammy-18 main" | sudo tee -a /etc/apt/sources.list.d/llvm-18.list
+           sudo apt-get update -qq
+           sudo apt-get install -y clang-format-18
      - name: Run clang-format
        run: |
          FILES=$(find . -type f ! -path '**/build/*' | egrep '\.(cc|cp|cpp|c++|cxx|cu|cuh|hh|hpp|hxx|hip|vert|frag)$')
-         clang-format -style=file -i $FILES
+         ${{env.CLANG_FORMAT}} -style=file -i $FILES
          git diff --exit-code
```
(new formatting errors on source code are reported when using the newer version)

Beanavil: Also, I'm not sure if this is the desired behaviour but currently `clang-format` is run for all files applicable, whether they were modified or not. If we want to run it only for the modified files (as the workflow is only triggered on `push` or `pull_request` on some branches) we could do:
```diff
env:
  CLANG_FORMAT: /usr/lib/llvm-18/bin/clang-format
+ GIT_CLANG_FORMAT: /usr/lib/llvm-18/bin/git-clang-format

jobs:
  ...

  source-formatting:
    ...
    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0
      ...
      - name: Run clang-format
        run: |
-           FILES=$(find . -type f ! -path '**/build/*' | egrep '\.(cc|cp|cpp|c++|cxx|cu|cuh|hh|hpp|hxx|hip|vert|frag)$')
-           ${{env.CLANG_FORMAT}} -style=file -i $FILES
-           git diff --exit-code
+           git config --global --add safe.directory $GITHUB_WORKSPACE
+           Scripts/CodeFormat/check_format.sh origin/`if [[ "${{ github.event_name }}" == "push" ]]; then echo "develop"; else echo "${{ github.base_ref }}"; fi` --binary ${{env.CLANG_FORMAT}}
```
(`GIT_CLANG_FORMAT` is for the `check_format.sh` to find the `git-clang-format` installed)

Scripts/CodeFormat/check_format.sh
```diff
@@ -38,18 +43,30 @@ finish () {
...
+
+echo "Running $CLANG_FORMAT --style=file -i $FILES"
+$CLANG_FORMAT --style=file -i $FILES
+
+# Check if any files were modified by clang-format
+MODIFIED_FILES=$(git diff --name-only --diff-filter=M $FILES)
+if [ -z "$MODIFIED_FILES" ]; then
+    echo "No formatting changes needed."
+    exit 0
+fi
```
Beanavil: Why remove the use of `git-clang-format`? I don't think this new version of the script checks the modified files by the branch only

Mustaballer: @Beanavil I removed `git-clang-format` due to conflicts with `clang-format` styles and standardized on using `clang-format`. The updated formatting script now checks staged files that are added, copied, or modified (ACM). If preferred, I can revert to using `git-clang-format`, but this was also likely require updating the `linting.yml` to use `git-clang-format` as well.

Edit: I reverted to using `git-clang-format`. After updating my `clang-format` on my local machine to version 18.1.8, which matches the CI, there were no conflicts as mentioned earlier.

.github/workflows/linting.yml
Mustaballer: The current behavior of running `clang-format` for all applicable files is intentional, as this is a CI action meant to ensure consistent formatting across the entire codebase.

Scripts/CodeFormat/check_format.sh
dgaliffiAMD: Thanks, @Beanavil. Are you using `clang-format-18` with your pre-commit hooks? We were not, so that probably explains why we were seeing different behaviour.

.github/workflows/linting.yml
dgaliffiAMD: I think scanning the entire codebase is okay for the status check. 
If we want to reuse the check_format.sh script to do that, then that may help for consistency. Also, we won't have to maintain two lists of extensions to scan.

Scripts/CodeFormat/check_format.sh
```diff
@@ -48,8 +48,8 @@ grep '^no modified files to format$\|^clang-format did not modify any files$' \
 # Dump formatting diff and signal failure
 printf \
 "\033[31m==== FORMATTING VIOLATIONS DETECTED ====\033[0m
-run '\033[33m%s --style=file %s %s\033[0m' to apply these formating changes\n\n" \
+run '\033[33m%s --style=file %s %s\033[0m' to apply these formatting changes\n\n" \
 "$GIT_CLANG_FORMAT" "$*" "$SOURCE_COMMIT"
 
 cat "$scratch"
-exit 1
+exit 1
```
dgaliffiAMD: Shell linting... Can you re-add the empty line at the end of the script, please?

Beanavil: Closing the conversation as this is fixed

Scripts/CodeFormat/check_format.sh
Beanavil: Well, we are using the 17 version as of now, but I wanted to bump it to 18. We could also change it for now to 17 (so that we all use the same version) and later bump it. The logic to use clang-format-17 within the workflow is the same as I suggested for 18, but replacing the version number.

Scripts/CodeFormat/check_format.sh
```diff
@@ -1,9 +1,20 @@
...
+        --all-files)
+            ALL_FILES=true
+            shift
+            ;;
+        *)
+            SOURCE_COMMIT="$1"
+            shift
+            ;;
+    esac
+done
```
Beanavil: Just a last comment here :). If this is a while loop then in the case when we don't specify the `--all-files` all the arguments are shifted (because we always go to the default case) and then we end up with no arguments to pass to `git-clang-format` and with `SOURCE_COMMIT` being the last argument specified, which is not the expected behaviour. As I see it, right now the logic should be:
- if all files should be formatted/checked, specify `--all-files` as the first argument
- if only the modified files should be formatted/checked, specify a base revision commit as the first argument (which was the previous behaviour)

So I think this while should be instead something along the lines of:
```suggestion
# Parse arguments
ALL_FILES=false
SOURCE_COMMIT=""
# Either a base SHA is used or all files are checked
if [[ $# -gt 0 ]]; then
    case "$1" in
        "--all-files")
            ALL_FILES=true
            ;;
        *)
            SOURCE_COMMIT="$1"
            ;;
    esac
    shift
fi
```
what do you think?

Mustaballer: @Beanavil Thank you for the detailed feedback! You are right; the while loop could indeed result in unintended behavior if --all-files is not specified. I have made the necessary adjustments in my recent commit👍

Beanavil: Awesome, thanks!

dgaliffiAMD (approved):
Thank you.

Beanavil (approved):

PR #142, opened 2024-07-16
Fix tests according to latest compiler changes

jtpatel:
Fix tests according to latest compiler changes


dgaliffiAMD:
Renaming to remove reference to internal ticket, SWDEV-468634. You can include this in the message body when it's time to "Squash and Merge".
As I understand, while this will fix compilation on rocm-6.2, it will break rocm-6.1. Is there a way to keep the develop branch compiling on both rocm-6.1 and rocm-6.2? For example, a pre-procesing variable to check?

dgaliffiAMD (approved):
Now that rocm-6.2 has been released, let's merge this change.

PR #147, opened 2024-08-02
Create a build test CI for the HIP Examples

dgaliffiAMD:
The scope of this change is limited to the HIP-Basic folder. 
There was not enough disk space on the basic GitHub Runners to install the full ROCm stack and run the Library examples.

### DIFF
--- a/.github/workflows/build_applications.yml
+++ b/.github/workflows/build_applications.yml
@@ -39,7 +39,7 @@ jobs:
                   apt-get -y install ./amdgpu-install_${{ env.AMDGPU_INSTALLER_VERSION }}_all.deb &&
                   apt-get update -qq &&
                   apt-get install -y \
-                    rocm-dev rocm-llvm rocm-llvm-dev \
+                    rocm-dev rocm-llvm \
                     rocrand-dev hiprand-dev \
                     rocprim-dev hipcub-dev
                   rm -rf /var/lib/apt/lists/*
--- a/.github/workflows/build_applications.yml
+++ b/.github/workflows/build_applications.yml
@@ -1,0 +1,55 @@
+name: Build Applications
+
+on:
+  push:
+    branches: [ develop, release/** ]
+    paths:
+      - 'Applications/**'
+      - '.github/workflows/**'
+  pull_request:
+    branches: [ develop, release/** ]
+    paths:
+      - 'Applications/**'
+      - '.github/workflows/**'
+
+env:
+    ROCM_VERSION: 6.2
+    AMDGPU_INSTALLER_VERSION: 6.2.60200-1
+
+jobs:
+    build:
+        name: "Build Applications Examples"
+        runs-on: ubuntu-latest
+        container:
+            image: ubuntu:22.04
+        steps:
+          - uses: actions/checkout@v4
+
+          - name: Install dependencies
+            run: |
+                  apt-get update -qq &&
+                  apt-get install -y build-essential g++ glslang-tools \
+                    python3 python3-pip libglfw3-dev libvulkan-dev locales wget
+                  python3 -m pip install --upgrade pip
+                  python3 -m pip install cmake
+          - name: Install ROCm Dev
+            run: |
+                  export DEBIAN_FRONTEND=noninteractive
+                  wget https://repo.radeon.com/amdgpu-install/${{ env.ROCM_VERSION }}/ubuntu/jammy/amdgpu-install_${{ env.AMDGPU_INSTALLER_VERSION }}_all.deb
+                  apt-get -y install ./amdgpu-install_${{ env.AMDGPU_INSTALLER_VERSION }}_all.deb &&
+                  apt-get update -qq &&
+                  apt-get install -y \
+                    rocm-dev rocm-llvm rocm-llvm-dev \
+                    rocrand-dev hiprand-dev \
+                    rocprim-dev hipcub-dev
+                  rm -rf /var/lib/apt/lists/*
+                  echo "/opt/rocm/bin" >> $GITHUB_PATH
+                  echo "ROCM_PATH=/opt/rocm" >> $GITHUB_ENV
+                  echo "LD_LIBRARY_PATH=/opt/rocm/lib:${LD_LIBRARY_PATH}" >> $GITHUB_ENV
+                  apt-get autoclean
+          - name: Configure and Build
+            shell: bash
+            run: |
+              cd Applications && mkdir build && cd build
+              cmake -DGPU_ARCHITECTURES=all -S ..
+              cmake --build . -j 4
--- a/.github/workflows/build_hip_basic.yml
+++ b/.github/workflows/build_hip_basic.yml
@@ -1,4 +1,4 @@
-name: Ubuntu 22.04
+name: Build HIP-Basic
 
 on:
   push:
--- a/.github/workflows/build_libraries.yml
+++ b/.github/workflows/build_libraries.yml
@@ -39,7 +39,7 @@ jobs:
                   apt-get -y install ./amdgpu-install_${{ env.AMDGPU_INSTALLER_VERSION }}_all.deb &&
                   apt-get update -qq &&
                   apt-get install -y \
-                    rocm-dev rocm-llvm rocm-llvm-dev \
+                    rocm-dev rocm-llvm \
                     rocrand-dev hiprand-dev \
                     rocprim-dev hipcub-dev \
                     rocblas-dev hipblas-dev \
--- a/.github/workflows/build_libraries.yml
+++ b/.github/workflows/build_libraries.yml
@@ -1,0 +1,60 @@
+name: Build Libraries
+
+on:
+  push:
+    branches: [ develop, release/** ]
+    paths:
+      - 'Libraries/**'
+      - '.github/workflows/**'
+  pull_request:
+    branches: [ develop, release/** ]
+    paths:
+      - 'Libraries/**'
+      - '.github/workflows/**'
+
+env:
+    ROCM_VERSION: 6.2
+    AMDGPU_INSTALLER_VERSION: 6.2.60200-1
+
+jobs:
+    build:
+        name: "Build Libraries Examples"
+        runs-on: ubuntu-latest
+        container:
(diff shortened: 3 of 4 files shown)


Beanavil:
Thanks for adding this! Perhaps we can have one workflow for each relevant folder (Applications, HIP-Basic and Libraries), that works fine [for me](https://github.com/Beanavil/rocm-examples/actions).

dgaliffiAMD:
> Thanks for adding this! Perhaps we can have one workflow for each relevant folder (Applications, HIP-Basic and Libraries), that works fine [for me](https://github.com/Beanavil/rocm-examples/actions).

That's amazing. Thanks for the suggestion, @Beanavil

.github/workflows/build_applications.yml
```diff
@@ -0,0 +1,55 @@
...
+                    python3 python3-pip libglfw3-dev libvulkan-dev locales wget
+                  python3 -m pip install --upgrade pip
+                  python3 -m pip install cmake
+          - name: Install ROCm Dev
+            run: |
+                  export DEBIAN_FRONTEND=noninteractive
+                  wget https://repo.radeon.com/amdgpu-install/${{ env.ROCM_VERSION }}/ubuntu/jammy/amdgpu-install_${{ env.AMDGPU_INSTALLER_VERSION }}_all.deb
+                  apt-get -y install ./amdgpu-install_${{ env.AMDGPU_INSTALLER_VERSION }}_all.deb &&
+                  apt-get update -qq &&
+                  apt-get install -y \
```
amd-isparry, line 41: Am I missing something? The point of the installer that you are getting in line 38 is to install rROCm, so why are you guessing what parts needs to be installed manually here?

.github/workflows/build_applications.yml
```diff
@@ -0,0 +1,55 @@
...
+        steps:
+          - uses: actions/checkout@v4
+
+          - name: Install dependencies
+            run: |
+                  apt-get update -qq &&
+                  apt-get install -y build-essential g++ glslang-tools \
+                    python3 python3-pip libglfw3-dev libvulkan-dev locales wget
+                  python3 -m pip install --upgrade pip
+                  python3 -m pip install cmake
```
amd-isparry, line 34: Things like cmake can be installed in various ways, in particular via the package manager and via pip. You should be prepared to handle errors if you try and install one way when the other way has previously been used.

.github/workflows/build_applications.yml
```diff
@@ -0,0 +1,55 @@
...
+                  echo "/opt/rocm/bin" >> $GITHUB_PATH
+                  echo "ROCM_PATH=/opt/rocm" >> $GITHUB_ENV
+                  echo "LD_LIBRARY_PATH=/opt/rocm/lib:${LD_LIBRARY_PATH}" >> $GITHUB_ENV
+                  apt-get autoclean
+          - name: Configure and Build
+            shell: bash
+            run: |
+              cd Applications && mkdir build && cd build
+              cmake -DGPU_ARCHITECTURES=all -S ..
+              cmake --build . -j 4
```
amd-isparry, line 55: Why 4? Why not 56 if you have a machine with 56 cores?

.github/workflows/build_hip_basic.yml
```diff
@@ -0,0 +1,53 @@
...
+      - '.github/workflows/**'
+
+env:
+    ROCM_VERSION: 6.2
+    AMDGPU_INSTALLER_VERSION: 6.2.60200-1
+
+jobs:
+    build:
+        name: "Build HIP Examples"
+        runs-on: ubuntu-latest
```
amd-isparry, line 22: Without me looking, I wonder how "ubuntu-latest" interacts with "ubuntu:22.04" a few lines further down. I think this needs a explanation. Perhaps a README file?

.github/workflows/build_hip_basic.yml
```diff
@@ -0,0 +1,53 @@
...
+        steps:
+          - uses: actions/checkout@v4
+
+          - name: Install dependencies
+            run: |
+                  apt-get update -qq &&
+                  apt-get install -y build-essential g++ glslang-tools \
+                    python3 python3-pip libglfw3-dev libvulkan-dev locales wget
+                  python3 -m pip install --upgrade pip
+                  python3 -m pip install cmake
```
amd-isparry, line 34: See previous comment on where you get cmake from.

.github/workflows/build_hip_basic.yml
```diff
@@ -0,0 +1,53 @@
...
+
+          - name: Install ROCm Dev
+            run: |
+                  export DEBIAN_FRONTEND=noninteractive
+                  wget https://repo.radeon.com/amdgpu-install/${{ env.ROCM_VERSION }}/ubuntu/jammy/amdgpu-install_${{ env.AMDGPU_INSTALLER_VERSION }}_all.deb
+                  apt-get -y install ./amdgpu-install_${{ env.AMDGPU_INSTALLER_VERSION }}_all.deb &&
+                  apt-get update -qq &&
+                  apt-get -y install rocm-dev rocm-llvm-dev
+                  echo "/opt/rocm/bin" >> $GITHUB_PATH
```
amd-isparry, line 43: /opt/rocm/bin might not exist, particularly if you are bypassing the installer.

.github/workflows/build_hip_basic.yml
```diff
@@ -0,0 +1,53 @@
...
+                  echo "ROCM_PATH=/opt/rocm" >> $GITHUB_ENV
```
amd-isparry, line 44: This should not be needed. If something fails when it is missing then it is a bug and should be reported. Consider this another test :-)

.github/workflows/build_hip_basic.yml
```diff
@@ -0,0 +1,53 @@
...
+                  echo "LD_LIBRARY_PATH=/opt/rocm/lib:${LD_LIBRARY_PATH}" >> $GITHUB_ENV
```
amd-isparry, line 45: This is wrong in many ways. It should not be needed, ROCm should be able to find its own libraries, /opt/rocm may not exist or it can point to the wrong place. If there is no existing LD_LIBRARY_PATH then you stick a spurious ":" on the end.

.github/workflows/build_hip_basic.yml
```diff
@@ -0,0 +1,53 @@
...
+                  apt-get autoclean
```
amd-isparry, line 46: Scripts should not be doing autoclean. The administrator of the machine may be keeping some package that was originally automatically installed but is no longer strictly required around. You have no right to delete that package.

.github/workflows/build_libraries.yml
```diff
@@ -0,0 +1,60 @@
...
+                    python3 python3-pip libglfw3-dev libvulkan-dev locales wget
+                  python3 -m pip install --upgrade pip
+                  python3 -m pip install cmake
+          - name: Install ROCm Dev
+            run: |
+                  export DEBIAN_FRONTEND=noninteractive
+                  wget https://repo.radeon.com/amdgpu-install/${{ env.ROCM_VERSION }}/ubuntu/jammy/amdgpu-install_${{ env.AMDGPU_INSTALLER_VERSION }}_all.deb
+                  apt-get -y install ./amdgpu-install_${{ env.AMDGPU_INSTALLER_VERSION }}_all.deb &&
+                  apt-get update -qq &&
+                  apt-get install -y \
```
amd-isparry, line 41: Installer the installer and then ignore it????

amd-isparry (commented):
I think this requires quite a lot of work.

.github/workflows/build_libraries.yml
dgaliffiAMD: The installer is being used to setup the repo files. 
This CIs are just running on the basic GitHub runners and I found that they were running out of disk space when trying to install all of rocm. Instead, we've had to take a more piecemeal approach.

.github/workflows/build_hip_basic.yml
dgaliffiAMD: This workflow is running in a docker image on a basic GitHub runner. This was done to save space. We were having issues installing all of rocm on these basic runners, due to lack of disk space.

.github/workflows/build_hip_basic.yml
dgaliffiAMD: Thanks. I give this a try.

.github/workflows/build_hip_basic.yml
dgaliffiAMD: Thanks. I give this a try.

.github/workflows/build_hip_basic.yml
dgaliffiAMD: But we're still using `apt install rocm-dev`. This will create it.

.github/workflows/build_applications.yml
dgaliffiAMD: Each run starts with a fresh docker image, so this shouldn't be an issue. You are correct, I could have used apt-get for cmake too. I think I chose pip because I'll generally get a newer version.

.github/workflows/build_hip_basic.yml
amd-isparry: /opt/rocm is managed by update-alternatives. This is needed for the  use case of wanting to have more than one version of ROCm installed at the same time on the same machine - think very large multi-user machines. /opt/rocm therefor may or may not point to the current version of ROCm that you are wanting to test.

If you are going to fall back on the "We are starting this in a fresh docker container so it will all be OK", then you need to spell out this assumption and that this code is of no value when testing the multi-version rocm use case. Otherwise it just looks like the code is wrong because you have missed this case.

.github/workflows/build_libraries.yml
amd-isparry: Fix the runners then! Are they self hosted runners?

dgaliffiAMD: Unfortunately, not. It is one of the basic runners.

.github/workflows/build_applications.yml
Beanavil: Given that the job uses ubuntu I think it can be done like so:
```suggestion
              cmake --build . -j `nproc`
```

PR #152, opened 2024-08-15
try running ctest to build ci tests

Mustaballer:
Run CTest after successful builds


Mustaballer:
Closed since CTests cannot pass since "no ROCm-capable device is detected" for CI

PR #155, opened 2024-08-27
Add "Tutorials" directory to SOURCE_DIRS in build_packages.sh

Mustaballer:
This PR adds the "Tutorials" directory to the `SOURCE_DIRS` array in the `build_packages.sh` script. This ensures that the content in the "Tutorials" directory is included during the package build process. 

Attached is the content of the source package file including the new Tutorial source files:
[source-with-tutorial-deb-pkg-contents.txt](https://github.com/user-attachments/files/16766382/source-with-tutorial-deb-pkg-contents.txt)

### DIFF
--- a/Scripts/Packaging/build_packages.sh
+++ b/Scripts/Packaging/build_packages.sh
@@ -52,6 +52,7 @@ SOURCE_DIRS=(
     "HIP-Basic"
     "Libraries"
     "LLVM_ASAN"
+    "Tutorials"
 )
 
 


Scripts/Packaging/build_packages.sh
```diff
@@ -307,9 +299,6 @@ pushd "$GIT_TOP_LEVEL" || exit
 # Print input variables
 print_input_variables
 
-# Build the project, including tests
-build_project
```
dgaliffiAMD: Isn't this what you are copying when you build the test packages?

Mustaballer: Yes @dgaliffiAMD, you are right—the build_project function produces the CTestTestfile.cmake files, which are then copied when building the test packages, I will revert the change

dgaliffiAMD (approved):

PR #163, opened 2024-09-16
Fix Tutorials config and build

Beanavil:
Got some build errors on our CI some time ago. Some of the them were fixed but in the end I needed to disable some build cases (most of Windows') because there wasn't any apparent fix for some of the issues:
- windows with ROCm because of some [linker issues with google benchmark](https://stackoverflow.com/questions/73494386/lnk2001-linker-error-while-linking-google-benchmark-lib)
- windows with CUDA for some toolkit versions (<=12.5) because of some nvcc issues with C++20 which seems to be solved in the [CUDA 12.5 Update 1](https://stackoverflow.com/questions/78586741/cudafe-died-with-status-0xc0000409-when-switching-to-c20-for-nvcc)


Beanavil:
@MKKnorr I think it should be reviewable now, all yours!

MKKnorr (approved):

PR #176, opened 2024-10-25
Fix hipSOLVER syevj_batched memory leak

MKKnorr:
Fix memory leak and move the allocation out of the loop


dgaliffiAMD (approved):
Looks good. Thank you.

PR #178, opened 2024-11-01
Update Docker build command

harkgill-amd:
Update docker build command to pass in render group ID as an argument. As discussed in https://github.com/ROCm/rocm-examples/pull/172


Beanavil (approved):
LGTM

PR #180, opened 2024-11-15
Add compiler check before running module_api and reduction tests

darren-amd:
This is a fix for compilation issues on RHEL and SUSE systems where the default GCC version does not support these examples.

Adding these checks for GCC version >= 10 before running the module_api example using "filesystem" and reduction example using "execution".


HIP-Basic/CMakeLists.txt
```diff
@@ -92,7 +92,9 @@ if(NOT "${GPU_RUNTIME}" STREQUAL "CUDA")
         endif()
     endif()
 
-    add_subdirectory(module_api)
+    if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "10.4")
+        add_subdirectory(module_api)
+    endif()
```
MKKnorr: This version check is specific to g++ and might not work for all c++ compilers.
Also the logic should be inverted - if the compiler is not g++ we add the target, otherwise we have to check. I.e.
```suggestion
    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "10.4")
    # g++ < 10.4 does not support all required C++17 features
    else()
        add_subdirectory(module_api)
    endif()
```

Also applies to the other check, but with C++20

darren-amd: Thanks! Implemented as suggested, also g++ <= 10.4 supports all features of C++20 necessary for the reduction examples, so made the minimum for those examples the same.

MKKnorr (approved):

PR #181, opened 2024-11-20
Fix rocsparse spitsv example residual check

MKKnorr:


Libraries/rocSPARSE/level_2/spitsv/main.cpp
```diff
@@ -211,7 +211,7 @@ int main()
     std::cout << "Iterations performed: " << iter_counter << std::endl;
 
     // Check last residual computed to confirm that convergence was successful.
-    const double last_residual = residual_history.back();
+    const double last_residual = residual_history[iter_counter - 1];
```
zichguan-amd: CI is failing because newer changes in rocSPARSE changed the way iteration counts are done, resulting in off-by-one error. Otherwise LGTM.

MKKnorr: Thanks, the documentation is not very specific about this. Fixed it. The other CI failures are known issues.

zichguan-amd (approved):
All LGTM, but I'm not from the dev team so you might want to wait for someone else's approval.

PR #189, opened 2025-01-28
Update README.md

peterClon:
HIP-Basic is at the top level, it is not under Libraries


MKKnorr (approved):
Right, thanks for fixing it!

PR #200, opened 2025-02-21
Fix using CUID to host-only compilation

zichguan-amd:
Remove CUID for host-only compilation since the tests do not expect CUID.


tcgu-amd (approved):
LGTM

PR #208, opened 2025-03-10
Update GH Actions for new branch names

danielsu-amd:
Sets `amd-staging` and `amd-mainline` branches as workflow triggers.
Update `build_applications` ROCm version to 6.3.3.


sohaibnd (approved):

danielsu-amd:
There are some errors when installing CUDA packages, should those be addressed in here? Maybe @tcgu-amd could take a look?

PR #247, opened 2025-06-13
[vcxproj][6.4][vs2022] Move projects from `6.2` to `6.4`

emankov:
+ [IMP] `HIP-VS 6.4` should be installed in `Visual Studio 2022` for building `ROCm-Examples` for both `AMD` and `NVIDIA` targets


SathiyarajRam (approved):
Approved for 6.4 SDK testing

PR #250, opened 2025-06-18
Fix using CUID to host-only compilation on Windows

zichguan-amd:
Add fix https://github.com/ROCm/rocm-examples/pull/200 to VS files


darren-amd (approved):
LGTM

PR #253, opened 2025-06-19
Update Windows CI

zichguan-amd:
- Check GUID first so it doesn't have to wait for HIP SDK to install
- Patch toolset version to use the latest public release version


darren-amd (approved):
LGTM

PR #254, opened 2025-06-24
SWDEV-539881 - Use std namespace for memset as it conflicts with HIPs'.

jtpatel:


zichguan-amd:
If it's not urgent, I'll close this one in favor of https://github.com/ROCm/rocm-examples/pull/235 since it already includes the fix. Just waiting on CI to be fixed to merge it.

PR #257, opened 2025-06-25
Fix shared object linking for examples that don't use hipcc

zichguan-amd:
Add rpath flag to link runtime SOs for examples that don't use hipcc. CMake adds it automatically but need to be explicitly specified in Makefiles. This way we don't need to have ROCm library path added to LD_LIBRARY_PATH.


darren-amd (approved):
LGTM

PR #259, opened 2025-07-04
Fix rocSPARSE csritilu0 example which was performing the convergence check incorrectly

jsandham:
- This fixes an issue that often passed but on some machines would regularly fail.
- Fix csritilu0 convergence check which was incorrectly grabbing the end of the convergence history array when it should grab the `2*iter-1` entry (where `iter` is the actual number of iterations that was required). The error is very easy to reproduce if the `d_data` array is explicitly filled with large numbers after it is allocated. I think this often passed because if `d_data` happened to be initialized with zeros, grabbing the end of the array would grab zero and so always pass as `0<tol`.

### DIFF
--- a/Libraries/rocSPARSE/preconditioner/csritilu0/main.cpp
+++ b/Libraries/rocSPARSE/preconditioner/csritilu0/main.cpp
@@ -81,7 +81,6 @@ int main()
     rocsparse_int* d_csr_col_ind{};
     double*        d_csr_val{};
     double*        d_LU{};
-    double*        d_data{};
 
     constexpr size_t size_csr_row_ptr = sizeof(*d_csr_row_ptr) * (n + 1);
     constexpr size_t size_csr_col_ind = sizeof(*d_csr_col_ind) * nnz;
@@ -90,13 +89,11 @@ int main()
     constexpr size_t size_LU          = sizeof(*d_LU) * length_LU;
     rocsparse_int    max_iter         = 100; /*maximum number of iterations*/
     const size_t     length_data      = 2 * max_iter;
-    const size_t     size_data        = sizeof(*d_data) * length_data;
 
     HIP_CHECK(hipMalloc(&d_csr_row_ptr, size_csr_row_ptr));
     HIP_CHECK(hipMalloc(&d_csr_col_ind, size_csr_col_ind));
     HIP_CHECK(hipMalloc(&d_csr_val, size_csr_val));
     HIP_CHECK(hipMalloc(&d_LU, size_LU));
-    HIP_CHECK(hipMalloc(&d_data, size_data));
 
     HIP_CHECK(
         hipMemcpy(d_csr_row_ptr, h_csr_row_ptr.data(), size_csr_row_ptr, hipMemcpyHostToDevice));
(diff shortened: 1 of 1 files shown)


Libraries/rocSPARSE/preconditioner/csritilu0/main.cpp
```diff
@@ -234,7 +234,8 @@ int main()
 
     // Check last residual computed to confirm that convergence was successful.
     HIP_CHECK(hipMemcpy(data.data(), d_data, size_data, hipMemcpyDeviceToHost));
-    const double last_residual       = data.back();
+
+    const double last_residual       = data[2 * max_iter - 1];
     const bool   csritilu0_converges = last_residual < tol;
```
YvanMokwinski: Also, this should be "<= tol"

Libraries/rocSPARSE/preconditioner/csritilu0/main.cpp
```diff
@@ -234,7 +234,8 @@ int main()
```
YvanMokwinski: std::vector\<double\> data(length_data); always initializes all the values to zero..

YvanMokwinski (commented):
Good catch, but  I feel there is something else going on.

Libraries/rocSPARSE/preconditioner/csritilu0/main.cpp
jsandham: It doesn't matter what `std::vector<double> data(length_data);` is initialized to. Whatever is there is always overwritten by the copy of `d_data` to the host.

jsandham:
> Good catch, but I feel there is something else going on.

The problem was not `std::vector` being initialized to zero or not (it is overwritten anyways by the copy of the device array to the host) but rather what entry we were accessing.

Libraries/rocSPARSE/preconditioner/csritilu0/main.cpp
YvanMokwinski: We only copy the relevant data:

https://github.com/ROCm/rocSPARSE/blob/develop/library/src/precond/itilu0/rocsparse_csritilu0_async_inplace.cpp#L1214

YvanMokwinski: clarified.

YvanMokwinski (approved):

jsandham:
@ROCm/rocm-examples-owners Can someone review when they have some time? We would like to get this merged as rocsparse CI runs the examples and the issue that this PR addresses causes our CI to fail.

zichguan-amd (approved):
LGTM

Libraries/rocSPARSE/preconditioner/csritilu0/main.cpp
```diff
@@ -224,18 +221,15 @@ int main()
...
-    std::vector<double> data(length_data);
+    std::vector<double> history(2 * max_iter);
+    ROCSPARSE_CHECK(
+        rocsparse_dcsritilu0_history(handle, alg, &max_iter, history.data(), buffer_size, temp_buffer));
 
-    // Check last residual computed to confirm that convergence was successful.
-    HIP_CHECK(hipMemcpy(data.data(), d_data, size_data, hipMemcpyDeviceToHost));
-    const double last_residual       = data.back();
-    const bool   csritilu0_converges = last_residual < tol;
+    const double last_residual       = history.back();
```
YvanMokwinski: = history[2*max_iter-1]

PR #269, opened 2025-07-16
Set default CMake build type to Release

zichguan-amd:
Avoids a compiler issue with rocfft_callback compiled with -O0


tcgu-amd (approved):
LGTM

PR #270, opened 2025-07-16
Update linting.yml

zichguan-amd:
Update cmake formatting check to include definitions from HipPlatform.cmake. Supresses unkown command warnings.


tcgu-amd (approved):
LGTM

PR #277, opened 2025-07-24
Fix LLVM binary lookup on Windows

### LINKED ISSUE #266: [Issue]: CMake lookup of LLVM binaries uses wrong path on Windows
### Problem Description

While looking at `HIP-Basic\CMakeLists.txt` I noticed the following lookup code for the LLVM binaries:

```cmake
    find_program(
        LLVM_DIS_COMMAND
        llvm-dis
        PATH_SUFFIXES bin
        PATHS ${ROCM_ROOT}/llvm ${CMAKE_INSTALL_PREFIX}/llvm
        NO_DEFAULT_PATH
    )
```

The issue here is that on Windows, `ROCM_ROOT` (which is set to the environment variable `HIP_PATH` which in turn is set by the HIP SDK installer) does not contain a `llvm` directory. Although the LLVM binaries are present, CMake won't find them and therefore disable the example on Windows.

### Operating System

Windows 11

### CPU

AMD Ryzen 7 PRO 8840U with Radeon 780M Graphics

### GPU

AMD Radeon 780M

### ROCm Version

ROCm 6.2 and ROCm 6.4

### ROCm Component

_No response_

### Steps to Reproduce

1. Install the official HIP SDK for Windows or use an AMD-internal version.
2. Observe that the installed HIP SDK version will not contain a `llvm` directory.
3. Configure the project with CMake and note the disabled examples.

### (Optional for Linux users) Output of /opt/rocm/bin/rocminfo --support

_No response_

### Additional Information

_No response_

j-stephan:
Fixes #266. As an additional bonus, the `assembly_to_executable` and `llvm_ir_to_executable` can now be built on Windows with the Ninja generator.

### DIFF
--- a/HIP-Basic/CMakeLists.txt
+++ b/HIP-Basic/CMakeLists.txt
@@ -31,12 +31,16 @@ if(CMAKE_SYSTEM_NAME MATCHES "Windows")
         CACHE PATH
         "Root directory of the ROCm installation"
     )
+    set(ROCM_LLVM_ROOT ${ROCM_ROOT})
+    set(PREFIX_LLVM_ROOT ${CMAKE_INSTALL_PREFIX})
 else()
     set(ROCM_ROOT
         "/opt/rocm"
         CACHE PATH
         "Root directory of the ROCm installation"
     )
+    set(ROCM_LLVM_ROOT "${ROCM_ROOT}/llvm")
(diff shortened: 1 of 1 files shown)


zichguan-amd (approved):
LGTM, thanks!

PR #284, opened 2025-07-31
Bump google test version to v1.12.0

zichguan-amd:
Support cmake 4 which is not compatible with cmake minimum < 3.5 used in google test 1.11.0


tcgu-amd (approved):
LGTM

PR #286, opened 2025-08-07
rocfft/hipfft: build callback example with -fgpu-rdc

evetsso:
Callback functions need this flag to be callable from device code in another translation unit, rather than building them into self-contained fully-linked binaries.

cf. https://clang.llvm.org/docs/HIPSupport.html#device-code-compilation


danielsu-amd:
Does this flag need to be added to the hipfft callback example as well?

evetsso:
hipFFT does not have a callback example here.  hipFFT's own repo does have one, but it hasn't been ported here.

danielsu-amd:
Bit confused, I'm referring to this one: https://github.com/ROCm/rocm-examples/tree/amd-staging/Libraries/hipFFT/callback

evetsso:
Ah I see, I'd based my branch on the stale develop branch.  I'll update this PR with the hipFFT callback changes too.

zichguan-amd (approved):

PR #288, opened 2025-08-11
Fix CUDA CI and move from self-hosted to GitHub-hosted runners 

zichguan-amd:
## Motivation

Don't update cuda inside the container since latest cuda 13 is not supported.
Move to GH-hosted to add more bandwidth.

## Technical Details

- Don't upgrade cuda
- Install `rocm-cmake` from apt since source code is private
- Remove make job limit
- Standardize directory change with `pushd/popd` and don't use `tmp` folder for cloning

## Submission Checklist

### DIFF
--- a/.github/workflows/build_applications_cuda.yml
+++ b/.github/workflows/build_applications_cuda.yml
@@ -19,7 +19,7 @@ env:
 jobs:
     build:
         name: "Build Applications CUDA"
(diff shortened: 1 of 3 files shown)


tcgu-amd (commented):
LGTM

huanrwan-amd (approved):
LGTM

PR #290, opened 2025-08-12
Explicitely link stdc++fs for GCC 8

zichguan-amd:
## Motivation

Fix fdtd compilation failure with GCC 8

## Technical Details

fdtd uses `std::filesystem` which requires explicit linking to `stdc++fs` when using GCC 8, see note at end of https://en.cppreference.com/w/cpp/filesystem.html.
`-lstdc++fs` is compatible with later GCC versions since they still ship `libstdc++fs`.
This does not cover use cases with `libc++` since ROCm LLVM is newer than 9.0.

## Test Result

Tested on EL 8.8 (Rocky), fdtd builds successfully with this PR, fails otherwise. (with both cmake and make)

## Submission Checklist

### DIFF
--- a/Applications/fdtd/CMakeLists.txt
+++ b/Applications/fdtd/CMakeLists.txt
@@ -63,7 +63,7 @@ endif()
 target_include_directories(${example_name} PRIVATE ${include_dirs})
 
 if(UNIX)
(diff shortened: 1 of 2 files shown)


lsudarsh-amd (approved):

tcgu-amd (commented):
LGTM

PR #307, opened 2025-09-05
Support experimental filesystem header for GCC versions in [5.3, 8.1)

zichguan-amd:
## Motivation

Fixes build issue with filesystem library on SLES 15.7 which comes with GCC 7.5 by default.

## Technical Details

Conditionally compile with available filesystem header using `__has_include`.
Skips examples in higher-level CMake if no filesystem header can be found.
Change `weakly_canonical` to `canonical` in `module_api` since `weakly_canonical` is not available in experimental/filesystem. https://en.cppreference.com/w/cpp/filesystem/canonical.html The function will error out if given a non-existing path, this does not change the example's functionality.
Skip reduction based on `execution` header availability instead of GCC version so it gets correctly skipped when using clang with older libstdc++.

## Test Plan

Compiles with GCC up to 5.3 https://godbolt.org/z/q5fofsbPd.
Builds `fdtd`, `linker_apis_file`, and `module_api` with CMake and make correctly on SLES 15.7 with GCC 7.5

## Submission Checklist


tcgu-amd (approved):
LGTM

PR #319, opened 2025-09-26
Add rocProfiler-SDK examples

adeljo-amd:
## Motivation

Add rocProfiler-SDK examples to ROCm examples. These examples were taken from the rocProfiler-SDK repository

## Technical Details

- Targets the latest API in rocProfiler-SDK from develop (which is not available in 6.4.3)

## Submission Checklist


zichguan-amd (approved):

Common/rocprofiler_utils.hpp
```diff
@@ -0,0 +1,388 @@
...
+        {                                                                                      \
+            std::cerr << "rocProfiler assertion failure: " << msg << " at " << __FILE__ << ':' \
+                      << __LINE__ << std::endl;                                                \
+            abort();                                                                           \
+        }                                                                                      \
+    }
+
+// Merged from Libraries/rocProfiler-SDK/common/filesystem.hpp
+#if !defined(ROCPROFILER_SAMPLES_HAS_GHC_LIB_FILESYSTEM)
+    #if defined __has_include
```
j-stephan, line 70: Is this required? I think we're using C++17 in the examples, so all compilers should know `__has_include`.

Common/rocprofiler_utils.hpp
```diff
@@ -0,0 +1,388 @@
...
+#include <random>
+
+#include <sstream>
+#include <stdexcept>
+#include <string>
+#include <string_view>
+#include <thread>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
```
j-stephan, line 56: Wow! Are all of those required for this file?

Libraries/rocProfiler-SDK/openmp_target/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+        "/opt/rocm"
+        CACHE PATH
+        "Root directory of the ROCm installation"
+    )
+endif()
+
+list(APPEND CMAKE_PREFIX_PATH "${ROCM_ROOT}")
+
+find_package(rocprofiler-sdk REQUIRED)
+find_package(rocprofiler-sdk-roctx REQUIRED)
```
j-stephan, line 52: Don't we need to find the OpenMP package here?

j-stephan (approved):
LGTM.

Common/rocprofiler_utils.hpp
adeljo-amd: Unfortunately yes

Libraries/rocProfiler-SDK/openmp_target/CMakeLists.txt
adeljo-amd: True, added

Common/rocprofiler_utils.hpp
adeljo-amd: Better to leave it in, although it may not be required.

bgopesh (commented):
I see some differences from the original samples which are mentioned here : https://github.com/ROCm/rocm-systems/blob/develop/projects/rocprofiler-sdk/samples/

Is it because these samples were ported long back or the changes are intentional?

bgopesh (commented):
Also please update main README for profiler examples.

adeljo-amd:
> I see some differences from the original samples which are mentioned here : https://github.com/ROCm/rocm-systems/blob/develop/projects/rocprofiler-sdk/samples/
> 
> Is it because these samples were ported long back or the changes are intentional?

The samples were taken from a while back, but besides some recent updates to a few of the samples, the rest should be the same in terms of logic. Can you a give an example of some of these differences? Most of the refactoring was to integrate the samples into the ROCm examples coding style and build system, but the logically they should be the same

>Also please update main README for profiler examples.

In what way? The README for the profiler examples follows the template set by the READMEs in the other library examples, but I'm open to any suggestions

bgopesh:
> > I see some differences from the original samples which are mentioned here : https://github.com/ROCm/rocm-systems/blob/develop/projects/rocprofiler-sdk/samples/
> > Is it because these samples were ported long back or the changes are intentional?
> 
> The samples were taken from a while back, but besides some recent updates to a few of the samples, the rest should be the same in terms of logic. Can you a give an example of some of these differences? Most of the refactoring was to integrate the samples into the ROCm examples coding style and build system, but the logically they should be the same
> 
> > Also please update main README for profiler examples.
> 
> In what way? The README for the profiler examples follows the template set by the READMEs in the other library examples, but I'm open to any suggestions

{3806E911-8751-431F-970C-B5D0A2040CA2}
I dont see any reference to profiler examples here unlike other modules.

bgopesh (approved):
Looks fine to me. Have left a comment to update reference of these examples in main README. Approving now.

adeljo-amd:
> > > I see some differences from the original samples which are mentioned here : https://github.com/ROCm/rocm-systems/blob/develop/projects/rocprofiler-sdk/samples/
> > > Is it because these samples were ported long back or the changes are intentional?
> > 
> > 
> > The samples were taken from a while back, but besides some recent updates to a few of the samples, the rest should be the same in terms of logic. Can you a give an example of some of these differences? Most of the refactoring was to integrate the samples into the ROCm examples coding style and build system, but the logically they should be the same
> > > Also please update main README for profiler examples.
> > 
> > 
> > In what way? The README for the profiler examples follows the template set by the READMEs in the other library examples, but I'm open to any suggestions
> 
> I dont see any reference to profiler examples here unlike other modules.

Capture23

Do you mean these ones? They're further down in the main readme

PR #326, opened 2025-10-08
Add rocJPEG examples

adeljo-amd:
## Motivation

Add rocJPEG examples to ROCm examples. These examples were taken from the rocJPEG repository

## Technical Details

- Targets the latest API in rocJPEG

## Submission Checklist


spolifroni-amd (approved):

adeljo-amd:
@zichguan-amd Seems the CI can't run the examples as it depends on VA-API JPEG decoder which appears to not be enabled in the CI environment. I think it'll be better to exclude the examples from the CI run

zichguan-amd:
Is it something we can easily enable on CI? If not we can exclude those from the tests.

adeljo-amd:
> Is it something we can easily enable on CI? If not we can exclude those from the tests.

Doesn't seem like it, the decoder seems to be only available when there's a display (from my tests). I've excluded the binaries from the tests

Libraries/rocJPEG/rocjpeg_decode/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+endif()
+
+list(APPEND CMAKE_PREFIX_PATH "${ROCM_ROOT}")
+
+find_package(rocjpeg REQUIRED)
+
+add_executable(${example_name} main.cpp)
+
+# Link to example library
+target_link_libraries(${example_name} PRIVATE rocjpeg::rocjpeg stdc++fs)
```
j-stephan: rocm-examples provides a lookup for the filesystem library (to guard against future compilers not shipping that anymore):

```suggestion
include("${CMAKE_CURRENT_LIST_DIR}/../../../Common/FindFilesystem.cmake")
target_link_libraries(${example_name} PRIVATE rocjpeg::rocjpeg ${CXX_FS_LIBRARY})
```

j-stephan: I don't see where `<filesystem>` is used in the code. Is it possible to not link against it at all?

Libraries/rocJPEG/rocjpeg_decode_batched/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+endif()
+
+list(APPEND CMAKE_PREFIX_PATH "${ROCM_ROOT}")
+
+find_package(rocjpeg REQUIRED)
+
+add_executable(${example_name} main.cpp)
+
+# Link to example library
+target_link_libraries(${example_name} PRIVATE rocjpeg::rocjpeg stdc++fs)
```
j-stephan: Same here.

Libraries/rocJPEG/rocjpeg_decode_perf/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+endif()
+
+list(APPEND CMAKE_PREFIX_PATH "${ROCM_ROOT}")
+
+find_package(rocjpeg REQUIRED)
+
+add_executable(${example_name} main.cpp)
+
+# Link to example library
+target_link_libraries(${example_name} PRIVATE rocjpeg::rocjpeg stdc++fs)
```
j-stephan: Same here.

Libraries/rocJPEG/rocjpeg_decode/CMakeLists.txt
zichguan-amd: `<filesystem>` is used in `Common/rocjpeg_utils.hpp` so that's fine, but would be nice to use `FindFilesystem.cmake`

zichguan-amd (commented):
LGTM. Just need to address the filesystem linking.
I tested on a headless system and was able to run the examples. I only had to install the `rocjpeg-dev` package. Can you verify that VA-API isn't just missing in the CI?

Libraries/rocJPEG/rocjpeg_decode/CMakeLists.txt
adeljo-amd: Fixed

adeljo-amd:
> LGTM. Just need to address the filesystem linking. I tested on a headless system and was able to run the examples. I only had to install the `rocjpeg-dev` package. Can you verify that VA-API isn't just missing in the CI?

Mmm, I'll need to check the CI scripts again - libva libraries are typically installed with rocjpeg-dev, and they should also be installed when running the rocJPEG script in Azure (even installing libva-amdgpu-dev should install all the other libva packages). Unless, the rocJPEG script only pulls the rocjpeg library files, and we need to add libva-amdgpu-dev to the CI manually

j-stephan (approved):

zichguan-amd (approved):

kiritigowda (approved):

PR #336, opened 2025-10-24
Bump hipBLASLt CMake minimum required version to 3.25.2

zichguan-amd:
## Motivation

Bump up the CMake version to match hipBLASLt's change in https://github.com/ROCm/rocm-libraries/commit/392aea29d39e1c61ac4ad1572e396c64738e1c95

## Submission Checklist

### DIFF
--- a/Libraries/CMakeLists.txt
+++ b/Libraries/CMakeLists.txt
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2022-2024 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (c) 2022-2025 Advanced Micro Devices, Inc. All rights reserved.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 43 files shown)


j-stephan (approved):
LGTM, but won't we need some sort of guard in hipBLASLt's parent directory in case someone tries to build this with an older version of CMake?

zichguan-amd:
You mean we should skip hipBLASLt if the version is not enough instead of failing the entire build?

j-stephan:
Yes. Otherwise, QA will discover some ancient version of RHEL we support and open a ticket for you ;-)

zichguan-amd:
Good point. I'll have a discussion with QA. I think we need to ask them to update the cmake version. This is not a HW support limitation, they need to be able to test hipBLASLt.

adeljo-amd (approved):
LGTM

zichguan-amd:
Added a check to skip hipBLASLt if CMake version is too low. Although QA needs to upgrade CMake in some distros to test hipBLASLt, users should be able to build other examples if they choose to not upgrade their CMake. This keeps the top-level CMake requirement as before rather than bumping the overall requirement.

j-stephan (approved):

PR #338, opened 2025-10-27
Fix VS sobel_filter input file path

zichguan-amd:
## Motivation

Fix input file path on Windows with Visual Studio.

## Technical Details

There's currently no way to get `__FILE__` to expand to a full path when building with Visual Studio using HIP clang. `GetFullPathNameA` get's the current working directory, so the example depends on where it gets invoked, and will fail if we are not in the root directory of rocm-examples.

Instead, we copy the input file to the output binary directory and build the input file path from the binary directory using `GetModuleFileNameA`.
Ref https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulefilenamea#parameters.

Building with CMake is not affected and works as expected ( `__FILE__` expands to the absolute full path).

## Submission Checklist

### DIFF
--- a/Applications/sobel_filter/main.hip
+++ b/Applications/sobel_filter/main.hip
@@ -315,20 +315,13 @@ int main(int argc, char* argv[])
 #ifdef _WIN32
     char separator = srcpath.find_last_of('/') == std::string::npos ? '\\' : '/';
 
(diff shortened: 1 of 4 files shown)


darren-amd (approved):
LGTM

PR #343, opened 2025-11-04
Add rocCV examples

adeljo-amd:
## Motivation

Add rocCV examples to ROCm examples. These examples were taken from the rocCV repository

## Technical Details

- Targets the latest API in rocCV

## Submission Checklist


adeljo-amd:
Seems like roccv isn't available as a package, currently investigating

j-stephan (approved):

zacharyvincze:
> Seems like roccv isn't available as a package, currently investigating

This is correct. The library is currently a source-only install.

zacharyvincze (approved):

adeljo-amd:
> > Seems like roccv isn't available as a package, currently investigating
> 
> This is correct. The library is currently a source-only install.

Thanks for the input. I have a question, do we want to include rocCV in the Azure CI? I've noticed that rocCV is not included in any of the Azure CI runners (not just for examples, but other runners), and would like to know if there's a reason why it's not been added there yet.

zichguan-amd (approved):
LGTM

kiritigowda (approved):

PR #348, opened 2025-11-05
Fix hipBLASLt library dependency in VS files

zichguan-amd:
## Motivation

Copy hipblaslt.dll to output directory for hipBLAS, rocBLAS, hipSOLVER, rocSOLVER examples since rocBLAS depends on hipBLASLt.
Do not copy rocsparse.dll since there's no rocSPARSE dependency for these examples.

## Technical Details

Solver libs seem to depend on rocSPARSE based on https://github.com/ROCm/ROCm/blob/develop/.azuredevops/components/hipSOLVER.yml#L53, but test passes without copying `rocsparse.dll` and there's no `rocsparse.dll` dependency from `roc/hipsolver.dll`.

## Test Plan

Build with `msbuild` and run executables without `HIPSDK_ROOT/bin` added to Path.

## Test Result

All affected examples run as expected, without needing `rocsparse.dll`.

## Submission Checklist

### DIFF
--- a/Libraries/hipBLAS/gemm_strided_batched/gemm_strided_batched_vs2017.vcxproj
+++ b/Libraries/hipBLAS/gemm_strided_batched/gemm_strided_batched_vs2017.vcxproj
@@ -32,6 +32,9 @@
     <Content Include="$(HIPExecutablePath)\rocblas.dll">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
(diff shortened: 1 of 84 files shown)


zichguan-amd:
rocSPARSE is an optional runtime dependency required by the Sp and Rf families of APIs. Currently examples do not contain these APIs so we can safely remove the `rocsparse.dll` copy.

darren-amd (approved):
LGTM

PR #357, opened 2025-11-12
Update rocWMMA CMake target name

zichguan-amd:
## Motivation

Fix Azure CI build failures following rocWMMA target name change to `roc::rocwmma` in https://github.com/ROCm/rocm-libraries/commit/8f545778b668a095f8154d69effb225db8962bc3

## Test Plan

See if it fixes Azure CI, might break GH workflows...

- [ ] Yes

- [ ] Yes

## Submission Checklist

### DIFF
--- a/Libraries/rocWMMA/CMakeLists.txt
+++ b/Libraries/rocWMMA/CMakeLists.txt
@@ -44,10 +44,19 @@ endif()
 
 list(APPEND CMAKE_PREFIX_PATH "${ROCM_ROOT}")
 
+find_package(hip)
+find_package(OpenMP)
 # Try to find rocWMMA as a package first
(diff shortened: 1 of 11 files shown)


adeljo-amd:
@zichguan-amd A simpler fix would be to omit the find_package command and just find the header manually/create library interface. That would avoid the target mismatch

zichguan-amd:
> @zichguan-amd A simpler fix would be to omit the find_package command and just find the header manually/create library interface. That would avoid the target mismatch

That's a good suggestion, but I think we should also demonstrate how to use `find_package`. It should be working now, forgot to find hip in the root CMake. This would support both the old and new naming.

adeljo-amd (approved):
LGTM

PR #378, opened 2025-12-16
fix the issue of GEMM validation failure

zhangnju:
## Motivation

when trying GEMM sample on R9700, if I change A_rows,A_cols,B_cols from default value to be 4096, validation will fail

## Technical Details

1) set the input matrix value to be random, which is the same with other GEMM application
2) develop a GEMM CPU validation function to test the GPU output , in order to speed up GEMM on CPU, I chose the block GEMM on CPU

## Test Plan

1) run the default value: the validation test can pass:
./hip_matrix_multiplication
Matrix multiplication: [2048x1024] * [1024x1024], block size: 16x16
Validation passed.
2) input size to be 4096, it still can pass 
 ./hip_matrix_multiplication --A_rows 4096 --A_cols 4096 --B_cols 4096
Matrix multiplication: [4096x4096] * [4096x4096], block size: 16x16
Validation passed.
3) run other size, and it still can pass
./hip_matrix_multiplication --A_rows 4096 --A_cols 512 --B_cols 2048
Matrix multiplication: [4096x512] * [512x2048], block size: 16x16
Validation passed.

## Test Result
all the test can pass 

- [ ] Yes
- [* ] No, does not apply to this PR.

- [ ] Yes
- [ *] No, does not apply to this PR.

## Submission Checklist

- [ *] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/HIP-Basic/matrix_multiplication/main.hip
+++ b/HIP-Basic/matrix_multiplication/main.hip
@@ -28,7 +28,6 @@
 #include <algorithm>
 #include <iostream>
 #include <vector>
-#include <vector>
 #include <cstdlib>
 #include <cassert>
 #include <cstddef>
--- a/HIP-Basic/matrix_multiplication/main.hip
+++ b/HIP-Basic/matrix_multiplication/main.hip
@@ -28,10 +28,9 @@
 #include <algorithm>
 #include <iostream>
 #include <vector>
-#include <cstdlib>
+
 #include <cassert>
 #include <cstddef>
-#include <memory>
 
 /// \brief Multiplies matrices \p A and \p B and stores the result to \p C.
 /// - The number of rows of the result matrix is equal to the number of rows of matrix A
@@ -214,9 +213,9 @@ int main(int argc, const char* argv[])
     std::vector<float> C(c_cols * c_rows);
 
     // Set matrix elements to random value on the host.
-    for (size_t i = 0; i < A.size(); ++i) A[i] = static_cast<float>(rand() / RAND_MAX );
-    for (size_t i = 0; i < B.size(); ++i) B[i] = static_cast<float>(rand() / RAND_MAX );
-
+    for (size_t i = 0; i < A.size(); ++i) A[i] = static_cast<float>(rand() / (RAND_MAX + 1.0f) );
+    for (size_t i = 0; i < B.size(); ++i) B[i] = static_cast<float>(rand() / (RAND_MAX + 1.0f) );
+    
     const size_t a_bytes = sizeof(float) * A.size();
     const size_t b_bytes = sizeof(float) * B.size();
     const size_t c_bytes = sizeof(float) * C.size();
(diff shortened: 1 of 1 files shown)


zichguan-amd:
cc @neon60 @j-stephan @adeljo-amd I wonder if this example overlaps with matrix multiplication from https://github.com/ROCm/rocm-examples/pull/375? If they are similar enough, we should probably just keep one.

zhangnju:
I think these two examples have different kernels ,which should not be redundant

HIP-Basic/matrix_multiplication/main.hip
```diff
@@ -165,11 +213,9 @@ int main(int argc, const char* argv[])
     std::vector<float> B(b_cols * b_rows);
     std::vector<float> C(c_cols * c_rows);
 
-    // Set matrix elements to a constant on the host.
-    std::fill(A.begin(), A.end(), 1.F);
-
-    constexpr float b_value = 0.02F;
-    std::fill(B.begin(), B.end(), b_value);
+    // Set matrix elements to random value on the host.
+    for (size_t i = 0; i < A.size(); ++i) A[i] = static_cast<float>(rand() / RAND_MAX );
```
zichguan-amd: Should be `static_cast<double>(rand()) / RAND_MAX`, `static_cast<float>(rand() / RAND_MAX )` would result in 0 most of the time (integer division)

HIP-Basic/matrix_multiplication/main.hip
```diff
@@ -165,11 +213,9 @@ int main(int argc, const char* argv[])
...
+    for (size_t i = 0; i < B.size(); ++i) B[i] = static_cast<float>(rand() / RAND_MAX );
```
zichguan-amd: Same as above

HIP-Basic/matrix_multiplication/main.hip
```diff
@@ -28,9 +28,10 @@
 #include <algorithm>
 #include <iostream>
 #include <vector>
-
+#include <cstdlib>
 #include <cassert>
 #include <cstddef>
+#include <memory>
```
zichguan-amd: Is this necessary?

zichguan-amd (changes requested):
I'm OK with the change, the CPU/GPU error should be in the same range, we can definitely use `double` for more precision. I'll let others weigh in.

HIP-Basic/matrix_multiplication/main.hip
zhangnju: Thanks for your reminding. we can change it to be "static_cast<float>(rand() / (RAND_MAX+1.0f) );", and it will generate [0,1) random float.

HIP-Basic/matrix_multiplication/main.hip
zhangnju: I have updated it to be "static_cast<float>(rand() / (RAND_MAX+1.0f) )", and verified that it can generate [0,1) random float value

HIP-Basic/matrix_multiplication/main.hip
zhangnju: I have removed them in the new commit

PR #381, opened 2025-12-18
Fix CMake return & update CMake messages

zichguan-amd:
## Motivation

Remove argument to `return()` in CMake. `return` does not take error code and will fail with CMP0140 enabled, it also does not terminate the file correctly under the old policy. Ref https://cmake.org/cmake/help/latest/command/return.html

Change "not building examples" message from STATUS to WARNING and make the preceding `find_package` quiet for better visibility.

## Technical Details

Standardize CMake warning and status messages:

- Use `WARNING` for missing dependencies & unsupported ASICs for better visibility
- Use `STATUS` for unsupported platforms (CUDA/Windows/Linux)

## Test Plan

rocJPEG is not properly skipped with `return(1)`. With this PR it is properly skipped.

- [ ] Yes

- [ ] Yes

## Submission Checklist


adeljo-amd (approved):
LGTM

PR #384, opened 2025-12-19
Fix Threads linking for rocProfiler-SDK examples

zichguan-amd:
## Motivation

Fixes build failure on RHEL 8.10

## Technical Details

See https://developers.redhat.com/articles/2021/12/17/why-glibc-234-removed-libpthread#
RHEL 8.10 has glibc 2.28 which requires extra linking to pthread

- [ ] Yes

- [ ] Yes

## Submission Checklist


adeljo-amd (approved):
LGTM

PR #387, opened 2026-01-08
Fix swizzle_tensor segfault

zichguan-amd:
## Motivation

Fixes `gemm_swizzle_a` segfault.

## Technical Details

The swizzle should be done on host, the tensors were allocated on device which gives a segfault when access on the host without dma.

## Test Plan

Local run with the PR

## Test Result

No segfault on gfx1100 (navi doesn't support swizzle yet so there's no solution)
```
./hipblaslt_gemm_swizzle_a 
Best solution time: 280.23 us (swizzle_a == 0)
No valid solution found!
No valid solution found!
F16 Swizzle Validation Error at index: 0, diff: 76
f8 Swizzle Validation Error at index: 0, diff: 76
```

- [ ] Yes

- [ ] Yes

## Submission Checklist


adeljo-amd (approved):
LGTM

PR #392, opened 2026-01-20
Add TheRock CI

zichguan-amd:
## Motivation

Update CI workflow to test with TheRock.

## Technical Details

Add build & test workflows for PR and commit with TheRock wheels and tarball:

- ubuntu:22.04 image
- gfx1100 & gfx1151

Add nightly runs for SLES 15.7
Build base docker image for ubuntu 22.04 and sles 15.7 to cut down CI time (~50%)

TODO:

1. Add nightly runs for other distros: RHEL 8
2. Add more runners for other ASICs: gfx12, gfx9, etc

## Test Plan

CI run.

## Test Result

Known test failures (component-side):

**rocprofv3-advanced (only with wheels):**
pmc counter collection issue

**rocprof-systems-basic and advanced:**
Need to run with `LD_LIBRARY_PATH="$ROCM_PATH/lib/rocprofiler-systems/:$ROCM_PATH/llvm/lib"`, but even with that both tests fail.

- [x] Yes, updated dockerfiles' README
- [] No, does not apply to this PR.

- [ ] Yes

## Submission Checklist


adeljo-amd:
LGTM so far, but for the current failures for gfx1100, those aren't something you're able to reproduce with the normal packages right?

zichguan-amd:
> LGTM so far, but for the current failures for gfx1100, those aren't something you're able to reproduce with the normal packages right?

Yea the failures aren't observed with 7.1.1 or 7.2. Only with TheRock.

zichguan-amd:
Need to disable the iGPU in the runner

adeljo-amd:
> > LGTM so far, but for the current failures for gfx1100, those aren't something you're able to reproduce with the normal packages right?
> 
> Yea the failures aren't observed with 7.1.1 or 7.2. Only with TheRock.

Okay, this seems like a good start so I'll approve after some of the CIs get closer to passing (e.g. the build libraries examples currently fails due to running out of space)

zichguan-amd:
Yup sounds good, that one needs to run either on self-hosted runner or should be fixed when we build a GA docker image, the GH-hosted runner is too small for current GA ROCm.

.github/workflows/therock_ci.yml
```diff
@@ -0,0 +1,223 @@
...
+          echo "HIP_PLATFORM=amd" >> $GITHUB_ENV
+          echo "HIP_DEVICE_LIB_PATH=$ROCM_PATH/lib/llvm/amdgcn/bitcode" >> $GITHUB_ENV
+          echo "$ROCM_PATH/bin" >> $GITHUB_PATH
+          echo "$ROCM_PATH/llvm/bin" >> $GITHUB_PATH
+
+      - name: Install TheRock tarball
+        if: ${{ matrix.install_method == 'tarball' }}
+        shell: bash
+        run: |
+          LATEST_TARBALL=$(curl -s https://rocm.nightlies.amd.com/tarball/ | grep -oE "therock-dist-linux-${{ matrix.gpu_config.therock_family }}-[0-9]+\.[0-9]+\.\w+\.tar\.gz" | tail -1)
```
sshi-amd: Might be good to have | sort -V | in here. 
```
LATEST_TARBALL=$(curl -s https://rocm.nightlies.amd.com/tarball/ \
  | grep -oE "therock-dist-linux-gfx110X-all-[0-9]+\.[0-9]+\.\w+\.tar\.gz" \
  | sort -V \
  | tail -1)
 ```

**Comment by sshi-amd** [comment]
I'm thinking maybe we could separate each test category? For example:
```
Start   1: applications_bitonic_sort
  1/277 Test   #1: applications_bitonic_sort ........................................   Passed    0.77 sec
        Start   2: applications_convolution
  2/277 Test   #2: applications_convolution .........................................   Passed    1.25 sec
        Start   3: applications_floyd_warshall
  3/277 Test   #3: applications_floyd_warshall ......................................   Passed    1.19 sec
        Start   4: applications_histogram
  4/277 Test   #4: applications_histogram ...........................................   Passed    1.17 sec
        Start   5: applications_prefix_sum
  5/277 Test   #5: applications_prefix_sum ..........................................   Passed    1.14 sec
        Start   6: applications_fdtd
  6/277 Test   #6: applications_fdtd ................................................   Passed    1.99 sec
        Start   7: applications_monte_carlo_pi
  7/277 Test   #7: applications_monte_carlo_pi ......................................   Passed    1.21 sec
        Start   8: applications_sobel_filter
  8/277 Test   #8: applications_sobel_filter ........................................   Passed    1.16 sec
  ``` 
  
  We can instead have like: 'applications all tests passed' 
  Maybe we don't have to replace this but just add it to the end as a addition instead and if an test case does fail we should link it?

.github/workflows/therock_ci.yml
zichguan-amd: Sounds like a good idea. If we don't want to rely on their default ordering, this should be the easiest way unless we do some processing with mtime. I'll add it.

zichguan-amd:
Do you mean for the summary? Currently there's not an easy way to separate the tests by category unless we do some filtering with the test names (which won't be enough because HIP-Doc and HIP-Basic all start with `hip_`).

sshi-amd:
I see, I feel like there merit to making a summary where it groups these tests together. Can we manually group these tests and test them in groups one at a time? It's obviously not a big issue but more appealing to the eyes. Furthermore I was also thinking of integrating github actions with Teams, maybe when tests fail it would automatically send a notification to a specific channel. LMK your thoughts on this.

zichguan-amd:
Would love the extra features, we can discuss this later

zichguan-amd:
@adeljo-amd @j-stephan only known failures remain. Can you give it a review? @sshi-amd will update the dockerfile CI and we can move to use custom-built image later.

j-stephan (approved):
LGTM

adeljo-amd (approved):
LGTM, did you report these issues to therock team (e.g. create a ticket)? It would be good to link these tickets to this PR (or ticket) for visibility purposes

zichguan-amd:
There are Jira tickets but no github issues for those failures

PR #402, opened 2026-01-30
Fix Vulkan validation errors

noahfredriks:
## Motivation
These were reported as driver errors, but they're indicative of application issues (minor in this case).

## Technical Details
In vulkan_interop, semaphore destruction was not waiting on the presentation engine to finish, which can be worked around by waiting for device idle instead of just submission fences. Vulkan_interop_mipmap was already doing this.

Both vulkan samples misused the wait semaphore provided to vkQueuePresentKHR by reusing it before the present completed. This solves that by creating enough semaphores to match the swapchain image count instead of matching the lower frame count as described in https://docs.vulkan.org/guide/latest/swapchain_semaphore_reuse.html

## Test Plan
vulkan_interop and vulkan_interop_mipmap were run on Radeon RX 9700 XT on Windows

## Test Result
Runs as before but without "Validation Error" messages.

- [ ] Yes

- [ ] Yes

## Submission Checklist

### DIFF
--- a/HIP-Basic/vulkan_interop/main.hip
+++ b/HIP-Basic/vulkan_interop/main.hip
@@ -598,8 +598,6 @@ struct frame
 
     /// The semaphore that guards the use of the swapchain image before it is ready.
     VkSemaphore image_acquired;
-    /// The semaphore that guards the present before the image is rendered.
-    VkSemaphore render_finished;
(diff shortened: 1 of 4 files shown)


zichguan-amd (approved):
Thanks so much for this fix! Tested on Windows gfx1201 with HIP SDK 7.1.1 and the Validation Errors are indeed gone.

PR #421, opened 2026-02-20
Update `LD_LIBRARY_PATH` for TheRock CI

j-stephan:
## Motivation

This should let the failing `rocprof-systems-{basic,advanced}` tests pass again.

## Technical Details

Updates `LD_LIBRARY_PATH` so the profiling tools can find their libraries.

The AMD SMI exception for the gfx1151 runners has been filed as bug AIPROFSYST-213 with the ROCm Systems Profiler team. Not much we can do there, but the tests should still pass.

## Test Plan

Tested locally.

## Test Result

Works.

- [ ] Yes

- [ ] Yes

## Submission Checklist


zichguan-amd (approved):
Looks like this also works for gfx1151 and fixes the rocprofv3-advanced test failing for gfx1100 wheels. It needed `LD_LIBRARY_PATH=${ROCM_PATH}/lib`.
LGTM thanks. With rccl disabled on gfx1151 we should have all tests green

PR #428, opened 2026-02-24
CI rework with reusable workflow file instead

sshi-amd:
## Motivation

Variation of https://github.com/ROCm/rocm-examples/pull/426

## Technical Details

Use of a reusable workflow file for the build process. Other workflows using this file can now define their own inputs, how they're invoked, and other customization.  

This variation requires the input labels to match the reusable workflows (gpu_config, install_method, distro) however the input choices do not. 

Another notable addition is that the distros choice must be an key in DISTRO_MAP in configure_ci.py, similar to GPU_CONFIG_MAP and INSTALL_METHODS. A novel nuance is the fact that the image urls in DISTRO_MAP also needs to match the uploaded docker image.

## Test Plan

Ran on self-hosted runner

## Test Result

https://github.com/sshi-amd/rocm-examples/actions/runs/22361448972

- [ ] Yes
- [ X] No, does not apply to this PR.

- [ ] Yes
- [X ] No, does not apply to this PR.

## Submission Checklist

- [ X] New examples contain proper CMake and Make files.
- [X ] I have updated relevant CI workflows and the new examples are being built and tested in CI.
- [X ] Check and guard against unsupported ASICs if relevant dependent libraries have limited support.
- [X ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


.github/build_tools/configure_ci.py
```diff
@@ -15,47 +15,63 @@
...
+# Distros to build against – keyed by short name.
+# Add new entries here to enable more distros (also add to workflow_dispatch options).
+DISTRO_MAP = {
+    "ubuntu-22.04": {"image": "ghcr.io/rocm/rocm-examples-ubuntu-22.04:latest", "label": "Ubuntu 22.04"},
+    "sles-15.7":    {"image": "ghcr.io/rocm/rocm-examples-sles-15.7:latest",    "label": "SLES 15.7"},
+}
+
+def _is_all(value):
+    """Return True when the input means 'use everything'."""
+    return not value or value == "all"
```
zichguan-amd, line 27: This triggers all combinations for automated workflows (empty values) with `if _is_all`, works fine for nightlies, but would be too much for PR. I think the logic here is fine, maybe we should pass down values from `ci_therock.yml` to limit what runs for PR

sshi-amd: It can be passed down via the input vars as seen in the latest [commit](https://github.com/ROCm/rocm-examples/pull/428/commits/6a7894f988e2e88f9b988ec391d7f408a2049db3) 

In this case we get only ubuntu running: https://github.com/sshi-amd/rocm-examples/actions/runs/22366322127/job/64733231566?pr=2

zichguan-amd (approved):

PR #438, opened 2026-03-11
Rhel 8 clean

sshi-amd:
## Motivation

Add RHEL 8 (EL8) support to the rocm-examples CI pipeline.

We use AlmaLinux 8 as the base image instead of official RHEL 8 because RHEL requires a Red Hat subscription for package access. Registering a subscription per container on every CI run is impractical. AlmaLinux 8 is a 1:1 binary-compatible RHEL 8 rebuild with the same glibc (2.28), kernel ABI, and package versions, so test results are representative of real RHEL 8 environments. This can be revisited if subscription-based images become feasible.

## Technical Details

Installs gcc-toolset-13 and sets it as the default compiler. This is required because RHEL 8's GCC 8 ships std::filesystem in a static library (libstdc++fs.a) whose ABI layout for path::_M_split_cmpts() is incompatible with TheRock's libhsa-runtime64.so (built with a newer GCC that has std::filesystem integrated into libstdc++.so). The HSA runtime calls _M_split_cmpts in a global constructor before main(), and the linker resolves it to the binary's GCC 8 version, causing an immediate segfault. GCC 13 from gcc-toolset-13 provides an ABI-compatible libstdc++_nonshared that resolves this.

Conditionally adds $ROCM_PATH/lib/rocm_sysdeps/lib to LD_LIBRARY_PATH and sets LIBVA_DRIVERS_PATH/LIBVA_DRIVER_NAME when the rocm_sysdeps/lib directory exists. This is needed because TheRock bundles its own VA-API driver (radeonsi_drv_video.so) in rocm_sysdeps/lib, and without these env vars, vaInitialize() fails for rocJPEG and rocDecode tests.

The condition (if [ -d ... ]) ensures Ubuntu and SLES workflows are unaffected — those distros use system VA-API packages and don't have rocm_sysdeps.

## Test Plan
Ran through CI and locally

## Test Result

https://github.com/sshi-amd/rocm-examples/actions/runs/22974154055

- [ ] Yes

- [ ] Yes

## Submission Checklist


zichguan-amd:
Alma8 is a good base image, we can remove the rhel8 ubi. I think is best to also discard rhel8 naming for alma8.

.github/build_tools/configure_ci.py
```diff
@@ -20,6 +20,7 @@
 DISTRO_MAP = {
     "ubuntu-22.04": {"image": "ghcr.io/rocm/rocm-examples-ubuntu-22.04:latest", "label": "Ubuntu 22.04"},
     "sles-15.7":    {"image": "ghcr.io/rocm/rocm-examples-sles-15.7:latest",    "label": "SLES 15.7"},
+    "rhel-8":       {"image": "ghcr.io/rocm/rocm-examples-rhel-8:latest",       "label": "RHEL 8"},
```
zichguan-amd: We can rename to "almalinux-8"

.github/workflows/build-rocm-examples-reusable.yml
```diff
@@ -72,7 +72,13 @@ jobs:
           echo "HIP_PLATFORM=amd" >> $GITHUB_ENV
           echo "HIP_DEVICE_LIB_PATH=${ROCM_PATH}/lib/llvm/amdgcn/bitcode" >> $GITHUB_ENV
           echo "PATH=${ROCM_PATH}/bin:${ROCM_PATH}/llvm/bin:${PATH}" >> $GITHUB_ENV
-          echo "LD_LIBRARY_PATH=${ROCM_PATH}/lib:${ROCM_PATH}/llvm/lib:${ROCM_PATH}/lib/rocprofiler-systems" >> $GITHUB_ENV
+          LD_LIB="${ROCM_PATH}/lib:${ROCM_PATH}/llvm/lib:${ROCM_PATH}/lib/rocprofiler-systems"
+          if [ -d "${ROCM_PATH}/lib/rocm_sysdeps/lib" ]; then
+            LD_LIB="${LD_LIB}:${ROCM_PATH}/lib/rocm_sysdeps/lib"
+            echo "LIBVA_DRIVERS_PATH=${ROCM_PATH}/lib/rocm_sysdeps/lib" >> $GITHUB_ENV
```
zichguan-amd: LIBVA_DRIVERS_PATH should not be needed if ROCM_PATH is set

.github/workflows/build-rocm-examples-reusable.yml
```diff
@@ -102,7 +108,13 @@ jobs:
...
```
zichguan-amd: Same as above

Dockerfiles/rhel-8-subscription.Dockerfile
```diff
@@ -0,0 +1,101 @@
+FROM registry.access.redhat.com/ubi8/ubi:latest
```
zichguan-amd: can be removed

Dockerfiles/almalinux-8.Dockerfile
```diff
@@ -0,0 +1,106 @@
+FROM almalinux:8
+
+ARG VULKAN_SDK_VERSION=1.4.335.0
+ARG GLFW_VERSION=3.4
+ARG GLSLANG_VERSION=13.1.1
```
zichguan-amd, line 5: Does almalinux 8 have distro packages for vulkan and glfw?

sshi-amd: Almalinux includes vulkan and glfw through the EPEL repo. Vulkan would be at version 1.3.283.0 and glfw would be 3.3. This differs from the SLES dockerfile which matches the current versions, I think its better to just build from source since its more explicit and easier to update.

.github/build_tools/configure_ci.py
```diff
@@ -20,6 +20,7 @@
 DISTRO_MAP = {
     "ubuntu-22.04": {"image": "ghcr.io/rocm/rocm-examples-ubuntu-22.04:latest", "label": "Ubuntu 22.04"},
     "sles-15.7":    {"image": "ghcr.io/rocm/rocm-examples-sles-15.7:latest",    "label": "SLES 15.7"},
+    "almalinux-8":       {"image": "ghcr.io/rocm/rocm-examples-almalinux-8:latest",       "label": "AlmaLinux 8"},
```
zichguan-amd: inconsistent formatting

.github/workflows/build-rocm-examples-reusable.yml
```diff
@@ -72,7 +72,11 @@ jobs:
           echo "HIP_PLATFORM=amd" >> $GITHUB_ENV
           echo "HIP_DEVICE_LIB_PATH=${ROCM_PATH}/lib/llvm/amdgcn/bitcode" >> $GITHUB_ENV
           echo "PATH=${ROCM_PATH}/bin:${ROCM_PATH}/llvm/bin:${PATH}" >> $GITHUB_ENV
-          echo "LD_LIBRARY_PATH=${ROCM_PATH}/lib:${ROCM_PATH}/llvm/lib:${ROCM_PATH}/lib/rocprofiler-systems" >> $GITHUB_ENV
+          LD_LIB="${ROCM_PATH}/lib:${ROCM_PATH}/llvm/lib:${ROCM_PATH}/lib/rocprofiler-systems"
+          if [ -d "${ROCM_PATH}/lib/rocm_sysdeps/lib" ]; then
+            LD_LIB="${LD_LIB}:${ROCM_PATH}/lib/rocm_sysdeps/lib"
+          fi
```
zichguan-amd: Is this still needed?

zichguan-amd (commented):
Otherwise LGTM

.github/workflows/build-rocm-examples-reusable.yml
sshi-amd: Looks like its not required in the latest CI run.

zichguan-amd (approved):

PR #440, opened 2026-03-20
Unify ROCM_INSTALL_DIR and ROCM_ROOT to ROCM_PATH

sshi-amd:
Standardize all ROCm installation path variables to use ROCM_PATH, aligning with TheRock infrastructure conventions.

Changes:
- Rename ROCM_ROOT to ROCM_PATH in all CMakeLists.txt files
- Rename ROCM_INSTALL_DIR to ROCM_PATH in all Makefiles
- Update build_packages.sh (variable, CLI flag --rocm-path)
- Update CI workflow to pass -DROCM_PATH instead of -DROCM_ROOT
- Update documentation references in README files
- Rename ROCM_ROOT_PATH C++ macro to ROCM_DEFAULT_PATH in rocWMMA hiprtc_gemm example (both CMake definition and usage)

Note: CMAKE_HIP_COMPILER_ROCM_ROOT in CMakePresets.json is a CMake built-in variable and is intentionally left unchanged.

Addresses: https://github.com/ROCm/rocm-examples/issues/429

## Test Plan

Tested through local CI

## Test Result

Passed CI: https://github.com/sshi-amd/rocm-examples/actions/runs/23312401667

- [x] Yes

- [ ] Yes

## Submission Checklist


zichguan-amd:
For CMake I was hoping to not have to set `-DROCM_PATH` if we can get it from the environment (I believe the env var is required for most ROCm component with TheRock). The priority for `ROCM_PATH` should be something like
1. Existing cache: `ROCM_PATH` already set (e.g. `-DROCM_PATH` or previous configure)
2. Unix: non-empty `$ENV{ROCM_PATH}`
3. Unix: non-empty `$ENV{HIP_PATH}`
4. Hardcoded default `/opt/rocm`

Would it be better if we extract this part of the CMake files into a separate file that we include to all the projects?
```
if(CMAKE_SYSTEM_NAME MATCHES "Windows")
    set(ROCM_ROOT
        "$ENV{HIP_PATH}"
        CACHE PATH
        "Root directory of the ROCm installation"
    )
else()
    set(ROCM_ROOT
        "/opt/rocm"
        CACHE PATH
        "Root directory of the ROCm installation"
    )
endif()

list(APPEND CMAKE_PREFIX_PATH "${ROCM_ROOT}")
```
Correct me if I'm wrong but I think it block is in every example.

zichguan-amd (approved):
LGTM thanks!

PR #441, opened 2026-03-24
Makefiles testing - CMake Source of Truth

sshi-amd:
## Motivation

Add makefiles testing

## Technical Details

CMake runs first and builds everything it can. A Python script (`find_cmake_built_dirs.py`) scans the CMake build tree for executables and matches them by name to `EXAMPLE :=` values in Makefiles. Only those Makefile directories are built. No `configure.sh`, no `config.mk`, no `require_deps.mk`.

- Use CMake as the single source of truth for dependency detection — no `configure.sh`, no `config.mk`, no `require_deps.mk`
- Add `find_cmake_built_dirs.py` — scans CMake build tree for executables, matches by name to `EXAMPLE :=` in Makefiles, outputs which Makefile directories to build
- All leaf Makefiles are unmodified from upstream (`amd-staging`) — no REQUIRED_DEPS additions
- Add `run_makefile_tests.sh` test runner with ctest-style output format
- Add `parse_ctest_results.py` to extract passed tests from ctest JUnit XML as the Makefile test allow list
- Workflow runs: CMake build → ctest → selective Makefile build → Makefile tests
- Upload both CMake and Makefile build artifacts and test logs
- Combined test summary (CMake + Makefile) on GitHub Actions summary tab
- Inline `$(shell pkg-config --libs ...)` in rocDecode Makefiles for FFmpeg `-L` paths on SLES

### Pros

- Zero new infrastructure in the Makefiles — all 29 leaf Makefiles are identical to upstream (`amd-staging`)
- No maintenance burden for future contributors — just write a Makefile with `EXAMPLE :=` matching the CMake target name (which is already the convention)
- Eliminates ~350 lines of configure/require_deps code
- Single source of truth for dependencies — if CMake can't build it, the Makefile won't be attempted
- No SLES edge cases to handle in a separate system

### Cons

- Makefile build depends on CMake running first — can't build Makefiles independently in CI
- Relies on the `EXAMPLE :=` naming convention matching CMake's `set(example_name ...)` — a mismatch means the Makefile silently won't be built (safe failure, but could be confusing)
- CMake build time is on the critical path — Makefile build can't start until CMake finishes
- Slightly less transparent — "why wasn't my Makefile built?" requires understanding the CMake matching logic

## Test Plan

Ran through local CI on GFX 1100: https://github.com/sshi-amd/rocm-examples/actions/runs/23504517478
(some tests were skipped due to time and ones that are unpredictable (rocprof-systems-advanced))

- [x] Yes

- [ ] Yes

## Submission Checklist


zichguan-amd:
closing in favor of https://github.com/ROCm/rocm-examples/pull/442

PR #443, opened 2026-04-02
Fix rocprof-systems-advanced: replace deprecated --rocm-domains with …

habajpai-amd:
…--enable-categories

## Motivation

The rocprof-systems-advanced test fails with Unrecognized command line option 'rocm-domains' when running against newer versions of rocprofiler-systems (e.g. from TheRock tarball), which have replaced the deprecated --rocm-domains CLI option with --enable-categories.

## Technical Details

Replaced --rocm-domains hip_api with --enable-categories rocm_hip_api in Tools/rocprof-systems/rocprof-systems-advanced.sh (line 62).

The --rocm-domains option was removed in newer rocprofiler-systems versions. The replacement --enable-categories uses a rocm_ prefix on category names (e.g. hip_api becomes rocm_hip_api). 

## Test Plan

Built rocm-examples against ROCm 7.13.0 with TheRock tarball with TheRock tarball on MI308 (gfx942) and ran the affected test:
cmake -S . -B build -DROCM_ROOT=$ROCM_PATH -DCMAKE_BUILD_RPATH=$ROCM_PATH/lib && cmake --build build -j
ctest -R rocprof-systems-advanced -VV

## Test Result

Test rocprof-systems-advanced passes after the fix. Previously failed with Unrecognized command line option 'rocm-domains'
image

- [ ] Yes

- [ ] Yes

## Submission Checklist

### DIFF
--- a/Tools/rocprof-systems/rocprof-systems-advanced.sh
+++ b/Tools/rocprof-systems/rocprof-systems-advanced.sh
@@ -59,7 +59,7 @@ echo "==========================================================================
 # Binary-rewrite our executable to include instrumentation points
 $EXAMPLE_INSTRUMENTER --output ${EXAMPLE_BIN}.inst -- $EXAMPLE_BIN
 # Restrict our experiment to HIP API calls, ignore everything else
-$EXAMPLE_RUNNER --profile --trace --rocm-domains hip_api -- ./${EXAMPLE_BIN}.inst
+$EXAMPLE_RUNNER --profile --trace --enable-categories rocm_hip_api -- ./${EXAMPLE_BIN}.inst
 
 echo "==============================================================================="
 echo "Instrumenting CPU performance counters"


zichguan-amd:
Thanks! LGTM, do you have the ticket/PR for the `rocm-domains` deprecation?

> Built rocm-examples against ROCm 7.1.2 with TheRock tarball 

Do you mean TheRock 7.12 or 7.13?

habajpai-amd:
> Thanks! LGTM, do you have the ticket/PR for the `rocm-domains` deprecation?
> 
> > Built rocm-examples against ROCm 7.1.2 with TheRock tarball
> 
> Do you mean TheRock 7.12 or 7.13?

The relevant PR is https://github.com/ROCm/rocm-systems/pull/3526 ("Refactor of rocprof-sys-avail tool"). It wrapped rocprofiler_sdk::config_settings() in an if(_init) guard in config.cpp, which deferred the registration of ROCPROFSYS_ROCM_DOMAINS from CLI parse time to library init time. Since Timemory auto-generates CLI flags from registered settings, --rocm-domains stopped being available as a CLI option. This landed between TheRock 7.12 (where the test passed) and 7.13 (where it fails).

And yes, I meant TheRock 7.13 -- built with 7.13.0a20260401 tarball. TheRock 7.12 (RC1, build 43) passes fine.

https://amd-hub.atlassian.net/browse/ROCM-20852 -- this is the jira ticket

Also, the two CI failures observed in this PR appear to be unrelated to the changes and seem to be pre-existing issues.

habajpai-amd:
> Thanks! LGTM, do you have the ticket/PR for the `rocm-domains` deprecation?
> 
> > Built rocm-examples against ROCm 7.1.2 with TheRock tarball
> 
> Do you mean TheRock 7.12 or 7.13?

TheRock 7.13

root@smc300x-clt-r4c3-34:/workspace/rocm-examples# amd-smi version
AMDSMI Tool: 26.3.0+96e30b429c | AMDSMI Library version: 26.3.0 | ROCm version: 7.13.0 | amdgpu version: 6.16.6 | hsmp version: N/A | AINIC version: N/A

zichguan-amd (dismissed):
Thanks LGTM

dgaliffiAMD (changes requested):
If you need a workaround, this will do, but it changes the functionality of the test. 
A better workaround option is to use `ROCPROFSYS_ROCM_DOMAINS=hip_api rocprof-sys-run ...`, which would be the same functionality, just setting the environment variable. 
Otherwise, we'll work a proper fix to the command-line arguments in rocprof-sys.

zichguan-amd:
> If you need a workaround, this will do, but it changes the functionality of the test. A better workaround option is to use `ROCPROFSYS_ROCM_DOMAINS=hip_api rocprof-sys-run ...`, which would be the same functionality, just setting the environment variable. Otherwise, we'll work a proper fix to the command-line arguments in rocprof-sys.

I'd prefer to keep the original functionality, can you elaborate on what `--enable-categories` differs from `--rocm-domains`? And is `ROCPROFSYS_ROCM_DOMAINS=hip_api` going to be the preferred method in the future?

dgaliffiAMD:
> > If you need a workaround, this will do, but it changes the functionality of the test. A better workaround option is to use `ROCPROFSYS_ROCM_DOMAINS=hip_api rocprof-sys-run ...`, which would be the same functionality, just setting the environment variable. Otherwise, we'll work a proper fix to the command-line arguments in rocprof-sys.
> 
> I'd prefer to keep the original functionality, can you elaborate on what `--enable-categories` differs from `--rocm-domains`? And is `ROCPROFSYS_ROCM_DOMAINS=hip_api` going to be the preferred method in the future?

Sorry, I'll elaborate. To the user, the functionality is _similar_. All the data that we collect is organized into "categories". If you use `--enable-categories rocm_hip_api`, then you are enabling the HIP API traces and that's it. (`rocprof-sys-run --help` to see a list of all the categories). The `--rocm-domain` option works more closely with the ROCProfiler-SDK domains. Setting `--rocm-domain hip_api` will affect the ROCm API tracing only, while other data (like hardware metric sampling) is still collected and recorded in your output. 

Which style should you use? Either is acceptable. This demonstrates that we have a gap in our test workflows, which favour environment variables and configuration files. However, we are closing that gap.

zichguan-amd:
Thanks for the detailed explanation. In that case I believe `--enable-categories rocm_hip_api` matches better the "I'm only interested in HIP APIs and nothing else" intent of the example.
https://github.com/ROCm/rocm-examples/blob/fd4b3a6e3dcd5910908e103a24b8728a9b3543bc/Tools/rocprof-systems/rocprof-systems-advanced.sh#L56-L63

habajpai-amd:
Thanks for the review feedback. I agree that fixing the root cause in rocprof-sys is the right approach rather than changing the test semantics here.

I've identified the regression -- PR #3526(https://github.com/ROCm/rocm-systems/pull/3526) added an if(_init) guard in config.cpp around rocprofiler_sdk::config_settings(), which defers registration of ROCPROFSYS_ROCM_DOMAINS past CLI parse time. Since Timemory auto-generates --rocm-domains from that setting, the flag disappears.

The fix is to remove the if(_init) guard so the setting is registered unconditionally during configure_settings(). rocprofiler_sdk::config_settings() only queries tracing domain names and registers Timemory settings -- no GPU initialization -- so it's safe to call during lightweight init. I'll put up a PR on rocm-systems for this.

Should I close this PR in favor of the upstream fix, or keep it open as a fallback using the ROCPROFSYS_ROCM_DOMAINS=hip_api environment variable approach @dgaliffiAMD  suggested?

habajpai-amd:
closing this one as it is now fixed in systems - https://github.com/ROCm/rocm-systems/pull/4684

PR #442, opened 2026-03-24
Makefile testing Add Makefile CI: configure-based dependency detection, build, and test

sshi-amd:
## Motivation

Mirror of #441 with a different implementation approach. Adds a Makefile build path to CI alongside the existing CMake/CTest flow, so the standalone `make` workflow stays exercised on every commit.

## Technical Details

A `Scripts/configure.sh` script probes the system for external libraries (FFmpeg, OpenCV, GLFW3, Vulkan, libdw, amd_comgr, glslangValidator) via `pkg-config` and link tests, then writes a `config.mk` at the repo root. Each leaf Makefile that has external dependencies declares them (`REQUIRED_DEPS := GLFW3`) and includes `Common/require_deps.mk`, which sets `SKIP_BUILD := 1` when any required dependency is missing. Examples without external dependencies are unaffected.

**This PR scope (PR1):** configure + Make build only. The Make-native test target and dependency-aware skip mechanism are deferred to a follow-up PR per review feedback.

### Added

- `Scripts/configure.sh` — pkg-config + link-test probe for FFmpeg, OpenCV, GLFW3, Vulkan, libdw, amd_comgr, glslangValidator. Writes `config.mk` at repo root.
- `Common/require_deps.mk` — leaf Makefiles declare `REQUIRED_DEPS`; build is skipped if any are missing. Has an inline pkg-config fallback so leaf Makefiles still work standalone without a prior `configure.sh` run.
- `Tools/ROCgdb/.gitignore` — was missing entirely, allowed `rocgdb-matmul` to slip into the tree.

### Modified

- ~390 leaf Makefiles: 28 of them declare `REQUIRED_DEPS` and wrap their build rules in `ifneq ($(SKIP_BUILD),1)`. The rest are unchanged in behaviour (touched only by the merge from upstream's `ROCM_PATH` rename).
- `Dockerfiles/almalinux-8.Dockerfile` — adds `pkgconf-pkg-config` (was missing, so every pkg-config probe failed on AlmaLinux 8) and extends `PKG_CONFIG_PATH` to include `/usr/local/lib(64)/pkgconfig` for source-built FFmpeg / GLFW.
- `.github/workflows/build-rocm-examples-reusable.yml` — adds a Makefile build step before CMake configure, plus separate "Collect Makefile build artifacts" / "Upload Makefile build artifacts" steps so Makefile-built binaries ship as their own artifact.
- `Makefile` (root) — `make configure` shortcut that calls `Scripts/configure.sh`.
[...]

PR #445, opened 2026-04-23
Add Make-native `test` target across all examples

sshi-amd:
## Motivation

Closes the testing half of the Make-build CI work that was deferred from #442. With this PR `make test` from the repo root builds and runs every leaf example that registered an `add_test` in CMake, matching the existing CMake/CTest coverage. CI just orchestrates; all test command/args/timeout logic lives in the Makefiles.

This PR is **stacked on top of #442** — its diff against `amd-staging` will look much smaller once #442 lands.

### Leaf Makefile `test:` target

Every leaf Makefile gains a `test:` target depending on its build target. Default invocation:

```makefile
.PHONY: test
test: $(EXAMPLE)
	./$(EXAMPLE) $(TEST_ARGS)
```

- `TEST_ARGS` is empty by default. `HIP-Doc/Tutorials/Programming-Patterns/bfs/Makefile` sets `TEST_ARGS := graph4096.txt` to mirror its CMake `add_test`.
- Leaves whose CMakeLists has no `add_test()` get an empty no-op `test:` (e.g. `HIP-Basic/{opengl_interop, vulkan_interop, static_device_library}`, `Libraries/MIVisionX/{canny, mv_objdetect, opencv_orb}`, three rocDecode examples that don't register).
- REQUIRED_DEPS leaves (~28) extend their existing SKIP_BUILD `else` no-op block with `test:` alongside `all:`/`clean:`.

### Parent Makefile cascade with per-parent skip list

Every parent Makefile (root + ~80 intermediates) declares `SKIP_FROM_TEST :=` and:

```makefile
test: TARGET=test
test: $(filter-out $(SKIP_FROM_TEST),$(EXAMPLES))
```

So `make test` cascades through the tree, skipping any subdirs the parent has explicitly excluded.

`Libraries/rocDecode/Makefile` populates `SKIP_FROM_TEST := $(EXAMPLES)`: all rocDecode examples need test data (videos under `$(ROCM_PATH)/share/rocdecode/`) that ctest currently guards behind `if(EXISTS …)` checks. Encoding that conditional logic in Make is doable but noisy; deferring to a follow-up. The leaf rocDecode Makefiles still have working `test:` targets so they can be run standalone via `cd Libraries/rocDecode/<example> && make test`.

### `Common/find_filesystem.mk` cleanup

The PR also converts the 19 remaining leaves with hard-coded `-lstdc++fs` to use the same `Common/find_filesystem.mk` probe introduced in #442 for the rocDecode leaves. After this, no leaf Makefile in the tree has a literal `-lstdc++fs`. Hygiene, not a bug fix — they currently work on all three CI distros.

### CI workflow

Adds `Makefile test` step after the existing test steps and `Upload Makefile test logs` artifact upload. Both gated with `if: !cancelled()` to mirror the rest of the test-side steps.

### Out of scope (potential follow-ups)

- **Per-GPU-arch test skipping** — `generate_skip_tests.py` is still used by the existing ctest step; PR2's per-directory `SKIP_FROM_TEST` is coarser.
- **Reduction tutorial gtest binaries** (`Tutorials/reduction/test/*.hip`) — no Makefile builds them today.
- **Wiring up rocDecode's conditional test-data probing** (CMake's `if(EXISTS ${ROCDECODE_TEST_VIDEO_HEVC})` pattern) — currently the entire rocDecode subtree is skipped at parent level.

## Test Plan

- Smoke-tested locally: `bitonic_sort` (bare run), `bfs` (with `TEST_ARGS`), root cascade dry-run reaches every leaf, `Libraries/rocDecode` cascade correctly does nothing.
- CI run after push: pending.

- [ ] Yes

- [ ] Yes

## Submission Checklist

- [x] New examples contain proper CMake and Make files. *(no new examples)*
- [x] Check and guard against unsupported ASICs if relevant dependent libraries have limited support. *(handled via per-parent `SKIP_FROM_TEST`)*
- [x] Look over the contributing guidelines at <https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests>.

### DIFF
--- a/.github/workflows/build-rocm-examples-reusable.yml
+++ b/.github/workflows/build-rocm-examples-reusable.yml
@@ -247,6 +247,18 @@ jobs:
           name: ctest-logs-${{ inputs.distro }}-${{ matrix.gpu_config.gpu_target }}-${{ steps.sanity-check.outputs.rocm_version || steps.install-tarball.outputs.rocm_version }}-${{ matrix.install_method }}
           path: build/Testing/Temporary/
 
+      - name: Makefile test
+        if: ${{ !cancelled() }}
+        run: |
+          make test 2>&1 | tee makefile_test_output.log
+
+      - name: Upload Makefile test logs
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: makefile-test-logs-${{ inputs.distro }}-${{ matrix.gpu_config.gpu_target }}-${{ steps.sanity-check.outputs.rocm_version || steps.install-tarball.outputs.rocm_version }}-${{ matrix.install_method }}
+          path: makefile_test_output.log
+
       - name: Clean the workspace
         if: always()
         run: find "$GITHUB_WORKSPACE" -mindepth 1 -delete
--- a/.github/workflows/build-rocm-examples-reusable.yml
+++ b/.github/workflows/build-rocm-examples-reusable.yml
@@ -250,7 +250,7 @@ jobs:
       - name: Makefile test
         if: ${{ !cancelled() }}
         run: |
-          make test 2>&1 | tee makefile_test_output.log
+          make test HIP_ARCHITECTURES="${{ matrix.gpu_config.gpu_target }}" 2>&1 | tee makefile_test_output.log
 
       - name: Upload Makefile test logs
         if: ${{ !cancelled() }}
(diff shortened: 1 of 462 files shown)


zichguan-amd (approved):
LGTM. TEST_ARGS might need to be aligned with ctest, we can iterate on it later.

PR #456, opened 2026-05-28
Fix nightly CI: remove redundant GCC/elfutils install steps, preserve LD_LIBRARY_PATH

mapatel-amd:
## Summary

The nightly CI was failing on all 4 SLES 15.7 jobs. Root cause: the workflow assumed a bare OS image, but the Docker images for both SLES 15.7 and Ubuntu 22.04 already ship with GCC 13 and elfutils pre-installed.

- The `Upgrade GCC (SLES)` step called `zypper addrepo leap-oss`, which exits code 4 because the alias already exists in the image — failing every SLES job before any build ran.
- The `Upgrade elfutils (SLES)` step built elfutils 0.186 from source, which is already installed to `/usr/local/lib` in the image.
- Both `Setup environment variables` steps overwrote `LD_LIBRARY_PATH` entirely, discarding the image's `/opt/vulkan-sdk/.../lib` path — causing `amdflang --help` to return exit code 255 in the wheel sanity check.
- The `Upgrade GCC (Ubuntu 22.04)` step installs gcc-13 from a PPA, but GCC 13.4.0 is already the default `gcc` in the Ubuntu image.

## Changes

- Remove `Upgrade GCC (SLES)` step
- Remove `Upgrade elfutils (SLES)` step
- Remove `Upgrade GCC (Ubuntu 22.04)` step
- Append to `LD_LIBRARY_PATH` (`:${LD_LIBRARY_PATH}`) in both `Setup environment variables` steps instead of overwriting it

## Test plan

- [x] Triggered `ci_nightly.yml` manually on this branch for `sles-15.7` + `gfx1100` + all install methods — **passed**
- [x] Triggered `ci_nightly.yml` manually on this branch for `sles-15.7` + `gfx1151` + all install methods — **passed**

### DIFF
--- a/.github/workflows/build-rocm-examples-reusable.yml
+++ b/.github/workflows/build-rocm-examples-reusable.yml
@@ -72,7 +72,7 @@ jobs:
           echo "HIP_PLATFORM=amd" >> $GITHUB_ENV
           echo "HIP_DEVICE_LIB_PATH=${ROCM_PATH}/lib/llvm/amdgcn/bitcode" >> $GITHUB_ENV
           echo "PATH=${ROCM_PATH}/bin:${ROCM_PATH}/llvm/bin:${PATH}" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


zichguan-amd (approved):
LGTM thanks

PR #460, opened 2026-06-08
Fix low-precision float CUDA guards

j-stephan:
## Motivation

The CUDA guard in the low precision floating-point examples' `CMakeLists.txt` is called before `ROCM_EXAMPLES_GPU_LANGUAGE` is defined. This will cause an error when building the examples individually.

## Technical Details

Moves the guard to the correct location.

## Test Plan

Tested locally.

## Test Result

Works.

- [ ] Yes

- [ ] Yes

## Submission Checklist

- [X] New examples contain proper CMake and Make files.
- [X] Check and guard against unsupported ASICs if relevant dependent libraries have limited support.


zichguan-amd (approved):
Nice catch LGTM thanks

PR #463, opened 2026-06-11
fix: skip the entire FFmpeg-dependent build

mapatel-amd:
**Summary of Changes**

All 8 rocDecode CMakeLists.txt files were modified to skip the entire FFmpeg-dependent build (not just tests) when running under TheRock.

**The Problem**

A CI log for a rocDecode example on a TheRock-based system showed this linker failure:

ld.lld: error: relocation R_X86_64_64 cannot be used against local symbol; recompile with -fPIC

The offending library: /usr/local/lib/libavcodec.a (TheRock's bundled FFmpeg).

**Root cause**: TheRock ships static FFmpeg libraries that were not compiled with -fPIC. The LLVM linker (lld) used by ROCm is stricter than GNU ld and
   hard-rejects position-dependent relocations in static libs when building a shared/position-independent target.

**What Changed**

**Before**: USING_THE_ROCK was detected, but only used to guard the add_test() calls at the bottom — the build itself still ran and tried to link against the non-PIC static FFmpeg, crashing lld.

**After**: The guard block is moved to before find_package(FFmpeg REQUIRED), with an early return(). If TheRock is detected (by presence of ${ROCM_PATH}/lib/rocm_sysdeps/lib), CMake skips the entire example — no FFmpeg find, no build, no tests.

Three files (video_decode_multi_files, video_decode_pic_files, video_to_sequence) had no guard at all before these changes.

The reference: [rocm-systems/3768](https://github.com/ROCm/rocm-systems/pull/3768) is the upstream TheRock PR that documents why this is necessary.

### DIFF
--- a/Libraries/rocDecode/video_decode/CMakeLists.txt
+++ b/Libraries/rocDecode/video_decode/CMakeLists.txt
@@ -46,6 +46,17 @@ include("../../../Common/ROCmPath.cmake")
 find_package(HIP REQUIRED)
 find_package(rocdecode REQUIRED)
 find_package(rocdecode-host 1.0.0 QUIET)
+# disable FFmpeg build and tests for TheRock, see https://github.com/ROCm/rocm-systems/pull/3768
+# TheRock ships static FFmpeg libs not compiled with -fPIC, which lld rejects
+set(USING_THE_ROCK OFF)
+if(EXISTS "${ROCM_PATH}/lib/rocm_sysdeps/lib")
+  set(USING_THE_ROCK ON)
+endif()
+if(USING_THE_ROCK)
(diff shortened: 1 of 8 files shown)


zichguan-amd:
Can you link the failed CI log? The FFmpeg dependencies should already be stripped with TheRock.  `/usr/local/lib/libavcodec.a` this looks more like system package then TheRock's.

mapatel-amd:
> Can you link the failed CI log? The FFmpeg dependencies should already be stripped with TheRock. `/usr/local/lib/libavcodec.a` this looks more like system package then TheRock's.

Nevermind, I think this problem came up when I was doing a fix for CDNA and misdiagnosed it. Disregard this.

PR #462, opened 2026-06-10
fix: add CDNA fallback to optical_flow kernels (ROCM-25855)

mapatel-amd:
## Summary

- CDNA GPUs (gfx942/gfx950, MI300/MI350 series) define `__HIP_NO_IMAGE_SUPPORT=1`, making the entire HIP texture API a compile-time error
- Add software bilinear sampler in `common.h` and `#if __HIP_NO_IMAGE_SUPPORT` dual paths in all four texture-using kernels
- RDNA path (hardware `tex2D` + `hipTextureObject_t`) is unchanged

## Test plan

- [x] Builds on RDNA (gfx1100/gfx1101) — confirmed by CI log from gfx1101 machine
- [x] CDNA testing completed on MI350X (gfx950)

PR #473, opened 2026-06-27
fix: guard HMM examples behind HSA_XNACK=1 in Makefile test

mapatel-amd:
Moved from #462 per review. HMM examples (hipMallocManaged/__managed__/hipMemPrefetchAsync) hang the gfx1151 Makefile test without HSA_XNACK=1 —
  hipDeviceSynchronize blocks on the faulted kernel. Guards them behind the XNACK conditional; only explicit_memory runs without it.


zichguan-amd:
Per Jan's comment: "I think you guys encountered a bug here. Unified memory should work even without HSA_XNACK. That's a performance optimization, not a requirement. If it hangs this should be filed as a bug for the HIP team.", we can temporarily disable these in CI, and file a ticket to HIP team.

mapatel-amd:
Sure let me get that out

mapatel-amd:
This branch isn't needed anymore.

PR #479, opened 2026-07-08
fix: make rocProfiler-SDK temporal ordering check non-fatal by default

mapatel-amd:
## Summary
- `external_correlation_id_request` aborted whenever a correlation ID appeared to retire before its operation's end timestamp.
- GPU dispatch and CPU-side retirement timestamps are in different clock domains; on RDNA3.5 APUs (gfx1151) they skew up to ~200 µs, so a correctly-ordered retirement
looks early and the example throws. gfx1100 (discrete) is unaffected.
- rocprofiler itself only warns and swaps these inversions by default, failing only under `ROCPROFILER_CI_STRICT_TIMESTAMPS`. This mirrors that policy: always print the
violation + count, only throw when that env var is set. The `unseen`/`unretired` invariants stay unconditional.

## Evidence
Nightly CI on gfx1151:
`terminate called ... what(): temporal ordering violation in correlation id retirement`
Observed skews: 1524 ns, 99057 ns, 209100 ns. Preceded by rocprofiler's own `hsa_amd_profiling_get_dispatch_time ... start time is after end time ... Swapping the values`
warning.
https://github.com/ROCm/rocm-examples/actions/runs/28764875296

## Test plan
- [x] Build the example
- [x] Run on gfx1151: exits 0, prints the tolerated warning
- [x] Run with `ROCPROFILER_CI_STRICT_TIMESTAMPS=1`: still aborts on violation

### DIFF
--- a/Libraries/rocProfiler-SDK/external_correlation_id_request/client.cpp
+++ b/Libraries/rocProfiler-SDK/external_correlation_id_request/client.cpp
@@ -582,10 +582,7 @@ tool_fini(void* tool_data)
     // rocprofiler uses), so genuine ordering bugs are still catchable in strict CI.
     if(temporal_ordering_violations > 0)
     {
-        const char* strict_ts_env = std::getenv("ROCPROFILER_CI_STRICT_TIMESTAMPS");
-        const bool  strict_timestamps
-            = (strict_ts_env != nullptr && strict_ts_env[0] != '\0' && strict_ts_env[0] != '0');
-        if(strict_timestamps)
+        if(common::get_env_bool("ROCPROFILER_CI_STRICT_TIMESTAMPS", false))
             throw std::runtime_error{"temporal ordering violation in correlation id retirement"};
         std::cerr << "warning: temporal ordering violations detected (tolerated; set "
                      "ROCPROFILER_CI_STRICT_TIMESTAMPS=1 to fail instead)\n"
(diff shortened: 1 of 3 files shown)


Libraries/rocProfiler-SDK/external_correlation_id_request/client.cpp
```diff
@@ -573,8 +573,24 @@ tool_fini(void* tool_data)
...
+    // correctly-ordered retirement look early; rocprofiler itself only warns and
+    // swaps such inversions by default. Mirror that policy: report the violation
+    // but only fail when ROCPROFILER_CI_STRICT_TIMESTAMPS is set (the same switch
+    // rocprofiler uses), so genuine ordering bugs are still catchable in strict CI.
     if(temporal_ordering_violations > 0)
-        throw std::runtime_error{"temporal ordering violation in correlation id retirement"};
+    {
+        const char* strict_ts_env = std::getenv("ROCPROFILER_CI_STRICT_TIMESTAMPS");
+        const bool  strict_timestamps
+            = (strict_ts_env != nullptr && strict_ts_env[0] != '\0' && strict_ts_env[0] != '0');
```
zichguan-amd: Maybe follow the same rules here and extract to a helper in Common utils? https://github.com/ROCm/rocm-systems/blob/6901b670c3e186f9c7052d6fc9748f0286536bdc/projects/rocprofiler-sdk/source/lib/common/environment.cpp#L93-L111, otherwise we could comment that this slightly differs from upstream behaviour.

zichguan-amd (approved):
LGTM with comments

Libraries/rocProfiler-SDK/external_correlation_id_request/client.cpp
mapatel-amd: Matching upstream behaviour now, will merge if passes.

mapatel-amd: Unrelated behaviour, merging.

PR #483, opened 2026-07-13
hipFile: Clean up after tests

j-stephan:
## Motivation

As reported by QA, the current CTest setup for the hipFile examples breaks when multiple users try to test them. Since the tests don't clean up after themselves in `/tmp`, name and permission collisions happen.

## Technical Details

Each user stores the hipFile test data in their own directory inside `/tmp`. Additionally, these directories are deleted after the test passes / fails.

- [ ] Yes

- [ ] Yes

## Submission Checklist

- [X] New examples contain proper CMake and Make files.
- [X] Check and guard against unsupported ASICs if relevant dependent libraries have limited support.


zichguan-amd (approved):
LGTM

PR #485, opened 2026-07-14
Add HIP-Basic execution context example

neon60:
## Motivation

Demonstrate partitioning a device's compute units into separate execution contexts, running a long-running background kernel and a latency-sensitive critical kernel on their own dedicated resources. Register the example in the HIP-Basic CMake build and the top-level README.

## Technical Details

Adds a new HIP-Basic/execution_context example (HIP's counterpart to CUDA green contexts).

- Queries the device's compute units with hipDeviceGetDevResource (hipDevResourceTypeSm), splits them with hipDevSmResourceSplit, wraps each group in a descriptor via hipDevResourceGenerateDesc, and creates an execution context per group with hipGreenCtxCreate.
A stream created on each context with hipExecutionCtxStreamCreate confines its kernels to that context's CUs.
- Runs a fixed critical workload concurrently with a device-saturating background kernel and measures the critical kernel's runtime. It first records a baseline where both kernels share all CUs, then sweeps several partition sizes (about an eighth, a quarter, and half of the device), printing a table of each configuration's critical- and background-kernel runtimes plus the critical kernel's speedup over the baseline. As the critical partition grows, its runtime drops below the contended baseline while the background kernel (confined to fewer CUs) takes longer.
- Partitioning is guarded by __HIP_PLATFORM_AMD__. On the CUDA backend, where the required runtime support may be unavailable, only the shared-CU baseline runs, and the CU-equivalent count is read from hipGetDeviceProperties.
- Includes a warm-up launch and hipDeviceSynchronize between cases so the device is idle before contexts are created and destroyed.
- Registered via add_subdirectory(execution_context) in HIP-Basic/CMakeLists.txt and listed in the top-level README.md.

Files added: main.hip, CMakeLists.txt, Makefile, README.md, .gitignore.

## Test Plan

- Build through the HIP-Basic CMake target hip_execution_context and run ./bin/hip_execution_context on an AMD GPU.
- Verify the baseline row plus one partitioned row per sweep size are printed, that the critical kernel's runtime falls (and
speedup rises) as the partition grows, and that the program completes cleanly.

## Test Result

Run on a 80-CU device (gfx1100): the critical kernel's runtime dropped from the shared baseline (~230 ms) to ~110 ms on larger partitions, roughly a 2x speedup, with the background-kernel runtime rising as its CU share shrank.

- [x] Yes

- [ ] Yes

## Submission Checklist


adeljo-amd (approved):
LGTM

j-stephan (approved):

HIP-Basic/execution_context/main.hip
```diff
@@ -0,0 +1,547 @@
...
+
+    std::cout << "Device ID: " << current_device << "\n";
+    std::cout << "Available CUs: " << cu_resources.sm.smCount << "\n";
+    std::cout << "Min. partition size: " << cu_resources.sm.minSmPartitionSize << "\n";
+    std::cout << "Co-scheduled alignment: " << cu_resources.sm.smCoscheduledAlignment << "\n";
+    // [read-cu-resource-end]
+
+    // [read-wq-config-start]
+    hipDevResource wq_config = {};
+    HIP_CHECK(hipDeviceGetDevResource(current_device, &wq_config, hipDevResourceTypeWorkqueueConfig));
```
zichguan-amd: This line fails the test on CI's gf1100 and 1151, does it require driver updates? Any ASIC support gap?

neon60: @zichguan-amd updated the code to not fail the test for this only log. I will ask HIP team about this.

zichguan-amd (approved):
LGTM

jujiang-del (approved):

saadrahim (approved):

PR #498, opened 2026-08-13
Fix FP8 round-trip example on non-native GPUs

skyguan92:
## Motivation

The FP8 example currently exits before launching its kernel unless the device is
`gfx94x`. Current HIP headers also provide host/device software conversion paths
when native FP8 conversion instructions are unavailable, so this excludes GPUs
that can execute the example correctly.

The round-trip validation also assigns raw `__hip_fp8_storage_t` bytes directly
to `float`. That compares the numeric value of the encoded byte instead of the
decoded FP8 value.

## Technical Details

- Decode FP8 storage through `__hip_cvt_fp8_to_halfraw` in a shared host/device
  helper.
- Remove the obsolete `gfx94x` early exit while retaining the existing FNUZ
  selection for `gfx94x` and OCP E4M3 elsewhere.
- Use `size_t` for the kernel index to remove the signed/unsigned warning.
- Document the decode helper and the portable fallback behavior.

This change enables correctness coverage; it does not claim native FP8 hardware
acceleration on architectures such as `gfx1100`.

## Test Plan

Test the exact commit on physical `gfx1100`, compare the unmodified baseline,
exercise both Make and CMake builds, validate the complete E4M3 raw-code space
plus a broad float corpus, trace the kernel dispatch, and compile the CI
architecture set.

## Test Result

Tested on two independently selected AMD Radeon Pro W7900D devices (GPU 0 and
GPU 7) with ROCm 7.14.0 / HIP 7.14.60850 in
`rocm/dev-ubuntu-24.04:7.14.0-full-amd64-digest-439edaa8`.

- Baseline: builds with a signed/unsigned warning, then prints `Need a gfx94x`
  and exits without launching the conversion kernel on both devices.
- Patched Make build: warning-free and reports `CPU and GPU round trip convert
  matches` on both devices.
- Patched CMake build: builds and reports the same passing result on both
  devices.
- Independent HIP validation: all 256 E4M3 raw codes and 65,557 deterministic
  float inputs match between host and device on both devices. The known-value
  oracle also passes: `1.1 -> 0x39 -> 1.125`.
- `rocprofv3 --kernel-trace` records one
  `float_to_fp8_to_float(..., unsigned long)` dispatch with a 32-thread workgroup.
- A `-Wall -Wextra -Werror` fat binary builds for the CI set: `gfx908`, `gfx90a`,
  `gfx942`, `gfx950`, `gfx1030`, `gfx1100`, `gfx1101`, `gfx1200`, and `gfx1201`.
- `clang-format` 18, `git diff --check`, and markdownlint pass.

- [x] Yes

- [ ] Yes

## Submission Checklist

- [x] Existing CMake and Make integrations build the updated example.
- [x] No CI workflow change is needed; the example already participates in the
  HIP-Doc build.
- [x] No unsupported-ASIC guard is needed because the HIP header provides the
  portable conversion path used here.
- [x] Reviewed the repository and ROCm pull-request guidelines.

### DIFF
--- a/HIP-Doc/Reference/Low-Precision-Floating-Point-Types/low_precision_float_fp8/README.md
+++ b/HIP-Doc/Reference/Low-Precision-Floating-Point-Types/low_precision_float_fp8/README.md
@@ -8,7 +8,7 @@ the
 
 ### Application flow
 
-1. Query the GPU device properties and check for a supported GPU architecture.
+1. Query the GPU device properties and select the FP8 interpretation for the GPU architecture.
 2. On the host side:
    1. Create an input vector and an output vector consisting of 32-bit floating-point numbers.
    2. Initialize the input vector.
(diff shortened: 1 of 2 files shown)


zichguan-amd:
@skyguan92 Thanks for contributing! @j-stephan can you validate the fix and verify if any docs changes are needed?

j-stephan (approved):
LGTM, I successfully tested the example on both Windows and Linux with RDNA3 GPUs. The functionality used here is already documented as well; we'll need to align the updated example and the documentation. I'll provide the necessary follow-up PRs. For now this can be merged as-is.

zichguan-amd (approved):

PR #91, opened 2024-01-25
Update README.md

Rmalavally:
Adding AI examples

PR #505, opened 2026-08-25
ci: hourly Quartz consumer workflow to test against latest ready nightly

zichguan-amd:
## Summary

Adds a standalone nightly workflow that tests rocm-examples against the **latest READY** ROCm nightly reported by [Quartz](https://github.com/ROCm/Quartz) (ROCm's CI/CD data hub), running **exactly once per new version**.

- **Gate:** resolves Quartz `release-nightly/latest.json` and requires the Linux **ROCm build** to be `success`. It deliberately does *not* gate on `overall_status`, which folds in every pipeline/phase and is routinely red from unrelated test failures.
- **Dedup:** a lookup-only `actions/cache` marker keyed on `(rocm_version, build_date)` ensures each version is tested once, not on every poll.
- **Fail-safe:** an in-progress build or an unreachable Quartz yields `resolved=false` and the run does nothing — doubling as a Quartz availability canary.

### Files

- `.github/quartz/read_status_json.py` — vendored **byte-for-byte** from ROCm/Quartz (`scripts/consumer`, schema v2) with a provenance header so it can be re-synced on schema bumps.
- `.github/quartz/consume_status.py` — the consumer/gate: an unsupported schema **major** is a hard failure (not a soft skip); requires `linux.rocm.build == success`; follows the `latest.json` symlink pointer; emits `resolved/rocm_version/build_date/source/wheels_url/tarballs_url` to `$GITHUB_OUTPUT`.
- `.github/quartz/tests/test_consume_status.py` — unit tests for the gate/resolve logic (good / not-ready / unavailable / bad-schema-major-exits / real-bug-propagates / absent platform).
- `.github/workflows/quartz_test.yml` — polls every 2h `cron "30 1-11/2 * * *"` (01:30–11:30 UTC, around the ~02:30 build-green time) + `workflow_dispatch` (`force`). Jobs: **gate** (consume + cache dedup + matrix), **test** (reusable build), **record** (marker on pass only), **report** (dynamic skip titles).
- `.github/workflows/build-rocm-examples-reusable.yml` — new optional `rocm_version` input that pins the exact nightly for both whl (`==<ver>`) and tarball (exact filename) installs, plus `wheels_url`/`tarballs_url` inputs so the install index/base come straight from Quartz's published URLs; all empty preserves today's latest-index behavior, so existing callers are unaffected.
- `.github/build_tools/configure_ci.py` — new opt-in `--exclude-preinstalled` flag that drops preinstalled-only images (the version-pinned stable image) from the matrix. Defaults off, so `ci_nightly.yml` is unaffected.

### Design decisions

- **Record on pass only** — a green run is the sole conclusive "tested." Any failure (transient infra *or* a real build/test break) leaves the version unrecorded so the next poll re-tests it, rather than masking it until the 7-day cache eviction. Retry is bounded: `resolve()` always prefers the newest ready build, so a genuinely broken nightly is superseded within the morning poll window, not retried hourly.
- **Matrix = full nightly minus preinstalled images** — 4 multi-arch distros × {gfx1100, gfx1151} × {whl, tarball} = 16 legs. The version-pinned stable image is excluded (via `configure_ci.py --exclude-preinstalled`) because its ROCm version is baked in and unrelated to the nightly we pin.
- **Poll window** — every 2h from 01:30–11:30 UTC. The Linux ROCm build starts ~00:02 and goes green ~02:24–03:09 on a normal night, so 01:30 sees it in progress (cheap skip), 03:30 usually catches it green, and the morning tail absorbs publish lag. A morning finish drops only the rare evening re-run, which self-heals when the next nightly supersedes it.
- **Cache eviction** — the 7-day marker eviction is accepted (a stale week means one harmless re-run).
- **`repository_dispatch`** (Quartz pushing on each build, removing polling+dedup) is deferred as a follow-up.

## Test plan

- [x] `consume_status.py` unit tests pass (`python3 -m unittest discover -s .github/quartz/tests`)
- [x] Consumer resolves `10.1.0a20260821` (build=success) against the live Quartz document
- [x] `configure_ci.py --exclude-preinstalled` drops the preinstalled image and keeps `distros`/`distro_map` in sync; both pinned tarballs return HTTP 200
- [x] Both workflow YAMLs parse
- [ ] `workflow_dispatch` run: confirm gate resolves, matrix pins the exact version, and `record` writes the marker
- [ ] Second immediate run shows the **skip** (already-tested) report

### DIFF
--- a/.github/build_tools/configure_ci.py
+++ b/.github/build_tools/configure_ci.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 """Configure CI matrix based on workflow inputs or defaults."""
+import argparse
 import os
 import json
 
@@ -40,9 +41,19 @@ def _is_all(value):
     return not value or value == "all"
 
 def main():
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--exclude-preinstalled",
+        action="store_true",
+        help="drop preinstalled-only images (e.g. the version-pinned stable image), "
+        "whose ROCm version is baked in and cannot be pinned to a nightly",
+    )
+    args = parser.parse_args()
+
     gpu_input = os.getenv("GPU_CONFIG", "")
     install_input = os.getenv("INSTALL_METHOD", "")
     distro_input = os.getenv("DISTRO", "")
+    exclude_preinstalled = args.exclude_preinstalled
 
     # Determine GPU configurations
     if _is_all(gpu_input):
@@ -78,11 +89,20 @@ def main():
     distro_map_out = {}
     for key in distro_keys:
         entry = DISTRO_MAP[key]
+        methods = entry.get("install_methods", install_methods)
+        # Skip preinstalled-only images (e.g. the version-pinned stable image) when asked:
+        # their ROCm version is baked in, so a caller pinning a specific nightly (the
+        # Quartz workflow) cannot use them.
+        if exclude_preinstalled and methods == [PREINSTALLED]:
+            continue
         distro_map_out[key] = {
             "image": entry["image"],
             "label": entry["label"],
-            "install_methods": entry.get("install_methods", install_methods),
+            "install_methods": methods,
         }
+
+    # Keep the distros list in sync with the (possibly filtered) map.
+    distro_keys = list(distro_map_out.keys())
 
     github_output = os.getenv("GITHUB_OUTPUT")
     if github_output:
--- a/.github/build_tools/configure_ci.py
+++ b/.github/build_tools/configure_ci.py
@@ -53,7 +53,6 @@ def main():
     gpu_input = os.getenv("GPU_CONFIG", "")
     install_input = os.getenv("INSTALL_METHOD", "")
     distro_input = os.getenv("DISTRO", "")
-    exclude_preinstalled = args.exclude_preinstalled
 
     # Determine GPU configurations
     if _is_all(gpu_input):
@@ -93,7 +92,7 @@ def main():
         # Skip preinstalled-only images (e.g. the version-pinned stable image) when asked:
         # their ROCm version is baked in, so a caller pinning a specific nightly (the
         # Quartz workflow) cannot use them.
-        if exclude_preinstalled and methods == [PREINSTALLED]:
+        if args.exclude_preinstalled and methods == [PREINSTALLED]:
             continue
         distro_map_out[key] = {
             "image": entry["image"],
--- a/.github/quartz/consume_status.py
+++ b/.github/quartz/consume_status.py
@@ -175,11 +175,20 @@ def main() -> None:
     rocm_version = result.status.rocm_version if result.status else ""
     build_date = result.status.build_date if result.status else ""
 
+    # Install URLs straight from Quartz (summary.linux.urls), so the reusable
+    # workflow pins the exact index/tarball base Quartz published rather than a
+    # hardcoded guess. Empty when there is no status document.
+    platform = result.status.platform(PLATFORM) if result.status else None
+    wheels_url = (platform.url("wheels") or "") if platform else ""
+    tarballs_url = (platform.url("tarballs") or "") if platform else ""
+
     set_github_outputs(
         resolved=str(result.resolved).lower(),
         rocm_version=rocm_version,
         build_date=build_date,
         source=result.source,
+        wheels_url=wheels_url,
+        tarballs_url=tarballs_url,
     )
 
     if result.resolved:
--- a/.github/quartz/consume_status.py
+++ b/.github/quartz/consume_status.py
@@ -1,0 +1,128 @@
+#!/usr/bin/env python3
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Resolve the latest ready ROCm nightly reported by Quartz.
+
+Adapted from ROCm/Quartz docs/status-json/example_consume_status.py for the
+hourly quartz_test workflow. It answers one question: is the latest nightly a
+build we can act on right now? A build qualifies when the Linux ROCm pipeline
+reports a successful build in Quartz's latest.json. When it does, this emits the
+exact version string to pin (rocm_version, e.g. 10.1.0a20260821); otherwise it
+emits resolved=false and the workflow does nothing that hour.
+
+"Latest when ready or nothing": only latest.json is consulted. There is no walk
+back to an older dated build -- testing a stale nightly is never what we want.
+
+Deduplication (test each version exactly once) lives in the workflow, not here:
+this script only reports what the latest build is, and the workflow gates on a
+cache marker keyed by (rocm_version, build_date).
+
+read_status_json.py is the sibling vendored reader. Running this file as
+`python3 .github/quartz/consume_status.py` puts this directory on sys.path[0],
+so the plain import below resolves it with no path shim.
+"""
+
+import argparse
+import os
+import sys
+
+from read_status_json import Status, StatusDocument, load_status
+
+# The build this workflow depends on: the Linux ROCm pipeline. Gate on this
+# specific platform + pipeline, never on overall_status (which folds in every
+# pipeline across platforms and is routinely red even when ROCm itself is fine).
+PLATFORM = "linux"
+PIPELINE = "rocm"
+
+
+def set_github_outputs(**outputs: str) -> None:
+    """Append step outputs to $GITHUB_OUTPUT; a no-op when run outside Actions."""
+    output_file = os.environ.get("GITHUB_OUTPUT")
+    if not output_file:
+        return
+    with open(output_file, "a") as handle:
+        for name, value in outputs.items():
+            handle.write(f"{name}={value}\n")
+
+
+def arch_is_good(status: StatusDocument, arch: str) -> bool:
+    """Return True if this build passed the gate we depend on.
+
+    - schema guard: the vendored reader is written against schema v2, so a
+      document whose schema_version does not start with "2." is treated as not
+      consumable rather than silently misread;
+    - the Linux ROCm build must be Status.success;
+    - if arch is non-empty, it must be one of the build's architectures (an
+      empty arch means a platform-level gate only).
+    """
+    if not status.schema_version.startswith("2."):
+        return False
+    platform = status.platform(PLATFORM)
+    if platform is None:
+        return False
+    if platform.pipeline_build_status(PIPELINE) != Status.success:
+        return False
+    if arch and arch not in platform.architectures:
+        return False
+    return True
+
+
+def resolve(source: str | None, arch: str):
+    """Load latest.json and decide whether it is ready to act on.
+
+    Returns (resolved, status, source_label). Every fetch is wrapped: an absent
+    or unreachable Quartz yields (False, None, "unavailable") so the workflow
+    simply skips the hour and doubles as a Quartz availability canary.
+    """
+    try:
+        status = load_status(source) if source else load_status()
+    except Exception as exc:  # noqa: BLE001 - fail-safe: never break the workflow
+        print(f"Quartz status unavailable: {exc}", file=sys.stderr)
+        return False, None, "unavailable"
+    if arch_is_good(status, arch):
+        return True, status, "latest"
+    return False, status, "not-ready"
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "arch",
+        nargs="?",
+        default="",
+        help="optional TheRock arch the build must include (e.g. gfx110X-all); "
+        "empty = platform-level gate only",
+    )
+    parser.add_argument(
+        "--latest-only",
+        action="store_true",
+        help="resolve latest.json only, with no fallback to an older dated build. "
+        "This is the only supported mode; the flag is accepted for explicitness.",
+    )
+    parser.add_argument(
+        "--source",
+        default=None,
+        help="override the status.json URL or path (default: Quartz latest nightly)",
+    )
+    args = parser.parse_args()
+
+    resolved, status, source = resolve(args.source, args.arch)
+    rocm_version = status.rocm_version if status else ""
+    build_date = status.build_date if status else ""
+
+    set_github_outputs(
+        resolved=str(resolved).lower(),
+        rocm_version=rocm_version,
+        build_date=build_date,
+        source=source,
+    )
+
+    if resolved:
+        print(f"resolved {rocm_version} ({build_date}) from {source}")
+    else:
+        print(f"no ready build ({source})")
+
+
+if __name__ == "__main__":
+    main()
--- a/.github/quartz/consume_status.py
+++ b/.github/quartz/consume_status.py
@@ -46,6 +46,11 @@ from read_status_json import (
 PLATFORM = "linux"
 PIPELINE = "rocm"
 
+# The status.json schema major this consumer understands. A bump to a new major
+# can move or rename the fields the accessors read, so an unsupported major is a
+# hard failure, not a soft skip (mirrors ROCm/Quartz example_consume_status.py).
+SUPPORTED_SCHEMA_MAJOR = "2."
+
 
 class Resolution(NamedTuple):
     """Outcome of resolve(): whether a ready build was found, the parsed status
@@ -70,15 +75,13 @@ def set_github_outputs(**outputs: str) -> None:
 def arch_is_good(status: StatusDocument, arch: str) -> bool:
     """Return True if this build passed the gate we depend on.
 
-    - schema guard: the vendored reader is written against schema v2, so a
-      document whose schema_version does not start with "2." is treated as not
-      consumable rather than silently misread;
     - the Linux ROCm build must be Status.success;
     - if arch is non-empty, it must be one of the build's architectures (an
       empty arch means a platform-level gate only).
+
+    The schema-major guard lives in resolve(): an unsupported major is a hard
+    failure there, not a soft "not ready" here.
     """
-    if not status.schema_version.startswith("2."):
-        return False
     platform = status.platform(PLATFORM)
     if platform is None:
         return False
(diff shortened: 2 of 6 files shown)


.github/build_tools/configure_ci.py
```diff
@@ -40,9 +41,19 @@ def _is_all(value):
...
+        action="store_true",
+        help="drop preinstalled-only images (e.g. the version-pinned stable image), "
+        "whose ROCm version is baked in and cannot be pinned to a nightly",
+    )
+    args = parser.parse_args()
+
     gpu_input = os.getenv("GPU_CONFIG", "")
     install_input = os.getenv("INSTALL_METHOD", "")
     distro_input = os.getenv("DISTRO", "")
+    exclude_preinstalled = args.exclude_preinstalled
```
HereThereBeDragons: this is a bit redundant. could just use args.exclude_preinstalled

.github/quartz/consume_status.py
```diff
@@ -0,0 +1,128 @@
...
+    simply skips the hour and doubles as a Quartz availability canary.
+    """
+    try:
+        status = load_status(source) if source else load_status()
+    except Exception as exc:  # noqa: BLE001 - fail-safe: never break the workflow
+        print(f"Quartz status unavailable: {exc}", file=sys.stderr)
+        return False, None, "unavailable"
+    if arch_is_good(status, arch):
+        return True, status, "latest"
+    return False, status, "not-ready"
```
HereThereBeDragons: not sure if you follow therock code style. comment from claude

resolve() returns a 3-field tuple (resolved, status, source). TheRock Python guide: "Tuple has 3+ fields → NamedTuple or dataclass." Also missing a return type annotation. Suggest:

```
class Resolution(NamedTuple):
    resolved: bool
    status: StatusDocument | None
    source: str
```

.github/quartz/consume_status.py
```diff
@@ -0,0 +1,128 @@
...
+    parser.add_argument(
+        "arch",
+        nargs="?",
+        default="",
+        help="optional TheRock arch the build must include (e.g. gfx110X-all); "
+        "empty = platform-level gate only",
+    )
+    parser.add_argument(
+        "--latest-only",
+        action="store_true",
```
HereThereBeDragons: we would also use here then BoolanOptionalAction

.github/quartz/consume_status.py
```diff
@@ -0,0 +1,128 @@
...
+        help="resolve latest.json only, with no fallback to an older dated build. "
+        "This is the only supported mode; the flag is accepted for explicitness.",
+    )
+    parser.add_argument(
+        "--source",
```
HereThereBeDragons, line 184: wondering if you want some exclusion between --latest-only and --source

.github/workflows/build-rocm-examples-reusable.yml
```diff
@@ -50,9 +55,15 @@ jobs:
...
           echo "## Target: ${{ matrix.gpu_config.gpu_target }} on ${{ fromJson(inputs.distro_map)[inputs.distro].label }}" >> $GITHUB_STEP_SUMMARY
           pip install --no-cache-dir --upgrade pip
+          PKG="rocm[libraries,devel,device-${{ matrix.gpu_config.gpu_target }}]"
+          # Pin to the exact nightly when a caller (e.g. the Quartz workflow) requests one;
+          # otherwise install whatever is latest on the index.
+          if [ -n "${{ inputs.rocm_version }}" ]; then
+            PKG="${PKG}==${{ inputs.rocm_version }}"
+          fi
           pip install --no-cache-dir \
             --index-url https://rocm.nightlies.amd.com/whl-multi-arch/ \
```
HereThereBeDragons: this URL is outdated.

new one is https://nightly.repo.amd.com/rocm/whl-next/ (aggregated index: rocm, pytorch, jax) or https://nightly.repo.amd.com/rocm/core/whl-next/ for rocm only


----
for quartz you can do:
summary.linux.urls.wheels will give you the correct url 
(use `def url(self, kind: str) -> str | None:` from read_status_json.py)

.github/workflows/build-rocm-examples-reusable.yml
```diff
@@ -72,9 +83,15 @@ jobs:
...
         run: |
-          TARBALL_LISTING=$(curl -s -H "Cache-Control: no-cache" -H "Pragma: no-cache" https://rocm.nightlies.amd.com/tarball-multi-arch/ \
-            | grep -oE "therock-dist-linux-${{ matrix.gpu_config.therock_family }}-tests-[0-9]+\.[0-9]+\.\w+\.tar\.gz")
-          LATEST_TARBALL=$(echo "${TARBALL_LISTING}" | sort -V | tail -1)
+          # Pin to the exact nightly when a caller (e.g. the Quartz workflow) requests one;
+          # otherwise pick the latest tarball on the index.
+          if [ -n "${{ inputs.rocm_version }}" ]; then
+            LATEST_TARBALL="therock-dist-linux-${{ matrix.gpu_config.therock_family }}-tests-${{ inputs.rocm_version }}.tar.gz"
+          else
+            TARBALL_LISTING=$(curl -s -H "Cache-Control: no-cache" -H "Pragma: no-cache" https://rocm.nightlies.amd.com/tarball-multi-arch/ \
```
HereThereBeDragons: same here: https://nightly.repo.amd.com/rocm/core/tarball/

and also same that if you want you can get the url and the tarball name created by quartz. but might be overkill or needs more rework in parallel to get it going

.github/workflows/build-rocm-examples-reusable.yml
```diff
@@ -72,9 +83,15 @@ jobs:
...
+              | grep -oE "therock-dist-linux-${{ matrix.gpu_config.therock_family }}-tests-[0-9]+\.[0-9]+\.\w+\.tar\.gz")
+            LATEST_TARBALL=$(echo "${TARBALL_LISTING}" | sort -V | tail -1)
+          fi
           mkdir /therock-tarball && cd /therock-tarball
           wget "https://rocm.nightlies.amd.com/tarball-multi-arch/${LATEST_TARBALL}"
```
HereThereBeDragons: https://nightly.repo.amd.com/rocm/core/tarball/

.github/workflows/quartz_test.yml
```diff
@@ -0,0 +1,165 @@
...
+      rocm_version: ${{ steps.consume.outputs.rocm_version }}
+      build_date: ${{ steps.consume.outputs.build_date }}
+      is_new: ${{ steps.dedup.outputs.is_new }}
+      distros: ${{ steps.config.outputs.distros }}
+      distro_map: ${{ steps.config.outputs.distro_map }}
+      gpu_configs: ${{ steps.config.outputs.gpu_configs }}
+      install_methods: ${{ steps.config.outputs.install_methods }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
HereThereBeDragons: pin to sha all checkouts

.github/workflows/quartz_test.yml
```diff
@@ -0,0 +1,165 @@
...
+
+      - name: Resolve latest ready nightly from Quartz
+        id: consume
+        run: python3 .github/quartz/consume_status.py --latest-only ""
```
HereThereBeDragons, line 77: i would have hoped that one could checkout quartz to get the scripts with a sparse checkout. but i also see the reasoning to copy them

wrong location but i mean the read_status_json

.github/workflows/quartz_test.yml
```diff
@@ -0,0 +1,165 @@
...
+
+      - name: Probe dedup marker
+        id: marker
+        if: steps.consume.outputs.resolved == 'true'
+        uses: actions/cache/restore@v4
```
HereThereBeDragons: pin

.github/workflows/quartz_test.yml
```diff
@@ -0,0 +1,165 @@
...
+        run: |
+          # is_new = there is a ready build we have not recorded yet (or --force).
+          if [ "${{ steps.consume.outputs.resolved }}" != "true" ]; then
+            echo "is_new=false" >> "$GITHUB_OUTPUT"
+          elif [ "${{ steps.marker.outputs.cache-hit }}" = "true" ] && [ "${{ github.event.inputs.force }}" != "true" ]; then
+            echo "is_new=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "is_new=true" >> "$GITHUB_OUTPUT"
+          fi
+
```
HereThereBeDragons, line 99: from my understanding the dedup caching part of  the `record:` step must follow here.
i dont know how long your test needs but you dont want to trigger a 2nd test run on the same build just because the first one is still in_progress

.github/workflows/quartz_test.yml
```diff
@@ -0,0 +1,165 @@
...
+      ${{ needs.gate.outputs.resolved == 'true'
+          && format('Skipped {0} (already tested)', needs.gate.outputs.rocm_version)
+          || 'No ready build yet' }}
+    steps:
+      - name: Report
+        run: |
+          if [ "${{ needs.gate.outputs.resolved }}" = "true" ]; then
+            echo "ROCm ${{ needs.gate.outputs.rocm_version }} (${{ needs.gate.outputs.build_date }}) already tested; skipping." >> "$GITHUB_STEP_SUMMARY"
+          else
+            echo "No ready ROCm nightly reported by Quartz yet." >> "$GITHUB_STEP_SUMMARY"
```
HereThereBeDragons: yet or already finished?

.github/workflows/quartz_test.yml
```diff
@@ -0,0 +1,165 @@
...
+    # Record on pass OR fail so a version is tested exactly once (a red result is a
+    # real regression signal, surfaced once, not retried hourly). Skip only if the
+    # test job was cancelled or never ran.
+    if: always() && needs.gate.outputs.is_new == 'true' && needs.test.result != 'cancelled' && needs.test.result != 'skipped'
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Write dedup marker
```
HereThereBeDragons, line 142: one comment from claude:

record writes the "tested" marker on any non-cancelled/non-skipped test result, including infra/setup failures (wheel index down, tarball 404), not just real regressions. That version is then never retried until 7-day cache eviction. Intended, or should setup failures skip the marker so the version retries next poll?

.github/build_tools/configure_ci.py
```diff
@@ -78,12 +89,21 @@ def main():
...
+            continue
         distro_map_out[key] = {
             "image": entry["image"],
             "label": entry["label"],
-            "install_methods": entry.get("install_methods", install_methods),
+            "install_methods": methods,
         }
 
+    # Keep the distros list in sync with the (possibly filtered) map.
+    distro_keys = list(distro_map_out.keys())
```
HereThereBeDragons, line 104: we also have rpm/deb support via quartz if interested?

.github/workflows/quartz_test.yml
```diff
@@ -0,0 +1,165 @@
...
+        run: touch .quartz-tested
+
+      - name: Save dedup marker
+        uses: actions/cache/save@v4
```
cgoea: pin to sha

.github/workflows/quartz_test.yml
zichguan-amd: We could definitely sparse checkout the read script, I don't have any preference.

.github/build_tools/configure_ci.py
zichguan-amd: Will definitly expand on that as a follow up

.github/workflows/quartz_test.yml
zichguan-amd: split the cases

.github/workflows/quartz_test.yml
zichguan-amd: This should prevent it, new jobs are queued and won't cancel the existing one, and when the running job finishes it would have cached the marker so the following jobs can exit accordingly.
```
# Never let a new poll cancel an in-flight test.
concurrency:
  group: quartz-test
  cancel-in-progress: false
```
but our jobs won't take long to finish anyway, usually ~30min, 2h interval is more than enough.

.github/workflows/quartz_test.yml
zichguan-amd: Now only mark on success

.github/quartz/consume_status.py
HereThereBeDragons, line 1: these are all good changes but i think this is something quartz should provide and fix in the read_status_json.py.

### Upstream Zichuan's #505 work into Quartz

#### Tier 1 — bug
- [ ] `scripts/consumer/read_status_json.py` — make `load_status()` follow the `latest.json` symlink (retry on `JSONDecodeError` -> fetch resolved target)
- [ ] `scripts/consumer/tests/read_status_json_test.py` — test: pointer body resolves; malformed JSON still raises

#### Tier 2 — robustness (in the example)
- [ ] `docs/status-json/example_consume_status.py` — wrap `load_status` in try/except (availability canary, emit `ready=false` not crash)
- [ ] `docs/status-json/example_consume_status.py` — schema-version guard (`"2."`)

#### Tier 3 — docs
- [ ] `docs/status-json/README.md` + `tutorial.md` — one line: `latest.json` is a symlink, `load_status` follows it

#### After it lands
- [ ] #505 drops `_read_pointer`/`load_latest`, calls `load_status` directly

.github/workflows/quartz_test.yml
```diff
@@ -133,14 +143,14 @@ jobs:
...
-        uses: actions/cache/save@v4
+        uses: actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
         with:
           path: .quartz-tested
           key: quartz-tested-${{ needs.gate.outputs.rocm_version }}-${{ needs.gate.outputs.build_date }}
 
       - name: Summary
         run: |
-          echo "Recorded ROCm ${{ needs.gate.outputs.rocm_version }} (${{ needs.gate.outputs.build_date }}) as tested; result=${{ needs.test.result }}." >> "$GITHUB_STEP_SUMMARY"
+          echo "Recorded ROCm ${{ needs.gate.outputs.rocm_version }} (${{ needs.gate.outputs.build_date }}) as tested (passed)." >> "$GITHUB_STEP_SUMMARY"
```
HereThereBeDragons, line 153: how is success defined here? just that build-rocm-examples-reusable ran independent of the build/test results? 
if not you might have multiple runs for the same rocm nightlies because rocm-examples has a bug and is legitly failing

.github/workflows/quartz_test.yml
```diff
@@ -116,14 +119,21 @@ jobs:
       gpu_configs: ${{ needs.gate.outputs.gpu_configs }}
       install_methods: ${{ needs.gate.outputs.install_methods }}
       rocm_version: ${{ needs.gate.outputs.rocm_version }}
+      wheels_url: ${{ needs.gate.outputs.wheels_url }}
+      tarballs_url: ${{ needs.gate.outputs.tarballs_url }}
     secrets: inherit
 
   record:
```
HereThereBeDragons, line 126: coming back to the discussion here where it belongs:

if you really only need 30min and check quartz in a 2h intervall i guess it will be ok.

but if you see any problems you probably want an additional  "in_progress" gate with a timestamp where you wait and if at that time the "successful-gate" is still from the day before and the "in_progress" gate is 4h+ old, you ignore it and trigger another build-rocm-examples-reusable.yml

HereThereBeDragons (commented):
looks good aside from the question about the gating.

do you want first the changes to land in quartz or you want to merge this here and then upgrade it at a later point? (also goes along with the question of the sparse checkout)

.github/quartz/consume_status.py
HereThereBeDragons: pr: https://github.com/ROCm/Quartz/pull/91

.github/workflows/quartz_test.yml
HereThereBeDragons: having some more thoughts about it in https://github.com/ROCm/Quartz/pull/91, the concurrency group should capture it. however in case downstream of the workflows you have some workflow_dispatch calls you will also need to guard them

.github/workflows/quartz_test.yml
```diff
@@ -0,0 +1,177 @@
...
+
+# Never let a new poll cancel an in-flight test.
+concurrency:
+  group: quartz-test
+  cancel-in-progress: false
+
+jobs:
+  gate:
+    name: "Resolve latest ready nightly"
+    runs-on: ubuntu-latest
```
HereThereBeDragons: ubuntu-24.04
for all runs-on.

.github/quartz/consume_status.py
```diff
@@ -0,0 +1,201 @@
...
+    """Return True if this build passed the gate we depend on.
+
+    - schema guard: the vendored reader is written against schema v2, so a
+      document whose schema_version does not start with "2." is treated as not
+      consumable rather than silently misread;
+    - the Linux ROCm build must be Status.success;
+    - if arch is non-empty, it must be one of the build's architectures (an
+      empty arch means a platform-level gate only).
+    """
+    if not status.schema_version.startswith("2."):
```
HereThereBeDragons: in quartz this will be now a hard failure if the major schema version is not matching

.github/quartz/consume_status.py
```diff
@@ -0,0 +1,201 @@
...
+def resolve(source: str | None, arch: str) -> Resolution:
+    """Load latest.json and decide whether it is ready to act on.
+
+    Every fetch is wrapped: an absent or unreachable Quartz yields
+    Resolution(False, None, "unavailable") so the workflow simply skips the poll
+    and doubles as a Quartz availability canary.
+    """
+    try:
+        status = load_latest(source)
+    except Exception as exc:  # noqa: BLE001 - fail-safe: never break the workflow
```
HereThereBeDragons: > #91 narrows to except (OSError, ValueError) specifically so a real bug (AttributeError, a rename in the reader) propagates instead of being mislabeled "Quartz unavailable" and retried forever. The broad catch collapses permanent programming errors into the transient bucket.

zichguan-amd: now mirrors the example

.github/quartz/consume_status.py
zichguan-amd: now mirrors the example

.github/workflows/quartz_test.yml
zichguan-amd: Conditioned on `needs.test.result == 'success'` which requires the entire matrix of `./.github/workflows/build-rocm-examples-reusable.yml` being green. So the ctests and Makefile testing must build and pass.

zichguan-amd:
Closing in favor of https://github.com/ROCm/rocm-examples/pull/512

PR #507, opened 2026-08-27
fix: Apply environment variable to hipthread examples

adeljo-amd:
## Motivation

Some of the hipthread examples fail to run on certain setups with a timeout on therock nightly builds. For now, the fix is to apply an environment variable to the failing examples.

## Technical Details

Apply environment variable to the ctest construction in the cmake files.

## Test Plan

Test on similar setups with therock nightly release.

## Test Result

Verified on similar setups that reported the issue locally and the fix works.

- [ ] Yes

- [ ] Yes

## Submission Checklist


marsavicc (approved):
LGTM, just one note:

`HIPTHREADS_VCORES_PER_WGP` is both compile time and runtime variable. When invoking ctest in your CI you could just set the env var there also.

Doing this via CMake is a viable way, but if you prefer doing it in runtime you can do it that way aslo.

zichguan-amd (approved):
LGTM, this should also be applied to the Makefile tests, maybe at compile time or in workflow would be better.

PR #512, opened 2026-09-02
ci: hourly Quartz consumer workflow to test against latest ready nightly (v2)

HereThereBeDragons:
## Summary

Adds a standalone nightly workflow that tests rocm-examples against the **latest READY** ROCm nightly reported by [Quartz](https://github.com/ROCm/Quartz) (ROCm's CI/CD data hub), running the expensive matrix **once per new version on success** — a passing version is recorded and skipped thereafter, while a failing one is re-tested on later polls until it passes or the next nightly supersedes it.

This is the sparse-checkout revision of #505: instead of vendoring Quartz's reader, the workflow reuses it directly from ROCm/Quartz at a pinned commit (unblocked by ROCm/Quartz#91). Co-authored with @zichguan-amd.

- **Gate:** resolves Quartz `release-nightly/latest.json` and requires the Linux **ROCm build** to be `success`. It deliberately does *not* gate on `overall_status`, which folds in every pipeline/phase and is routinely red from unrelated test failures.
- **Dedup:** a lookup-only `actions/cache` marker keyed on `(rocm_version, build_date)`, written **only on a green run**, skips a version once it has passed; a failing version stays unmarked and is re-tested on later polls (bounded — see *Record on pass only*).
- **Fail-safe:** an in-progress build or an unreachable Quartz yields `resolved=false` and the run does nothing — doubling as a Quartz availability canary.

### Files

- `read_status_json.py` is **not vendored**. The `gate` job sparse-checks out the single file from ROCm/Quartz (`scripts/consumer/read_status_json.py`) at pinned commit `910cc21e` (the ROCm/Quartz#91 merge, which ships the schema-v2 reader that follows the `latest.json` symlink) and puts it on `PYTHONPATH`. Bump the SHA to pick up reader fixes or a schema-major update.
- `.github/quartz/consume_status.py` — the consumer/gate: imports Quartz's reader (via `PYTHONPATH`) and calls `load_status`, which follows the `latest.json` symlink pointer internally; an unsupported schema **major** is a hard failure (not a soft skip); requires `linux.rocm.build == success`; emits `resolved/rocm_version/build_date/source/wheels_url/tarballs_url` to `$GITHUB_OUTPUT`.
- `.github/quartz/tests/test_consume_status.py` — unit tests for the gate/resolve logic (good / not-ready / unavailable / bad-schema-major-exits / real-bug-propagates / absent platform). Requires `PYTHONPATH` pointed at a Quartz checkout (see the module docstring: `git clone --depth 1 https://github.com/ROCm/Quartz /tmp/quartz`).
- `.github/workflows/quartz_test.yml` — polls every 2h `cron "30 1-11/2 * * *"` (01:30–11:30 UTC, around the ~02:30 build-green time) + `workflow_dispatch` (`force`). Jobs: **gate** (sparse-checkout reader + consume + cache dedup + matrix), **test** (reusable build), **record** (marker on pass only), **report** (dynamic skip titles).
- `.github/workflows/build-rocm-examples-reusable.yml` — new optional `rocm_version` input that pins the exact nightly for both whl (`==<ver>`) and tarball (exact filename) installs, plus `wheels_url`/`tarballs_url` inputs so the install index/base come straight from Quartz's published URLs; all empty preserves today's latest-index behavior, so existing callers are unaffected.
- `.github/build_tools/configure_ci.py` — new opt-in `--exclude-preinstalled` flag that drops preinstalled-only images (the version-pinned stable image) from the matrix. Defaults off, so `ci_nightly.yml` is unaffected.

### Design decisions

- **Reuse the Quartz reader via sparse checkout, don't vendor it** — a single-file sparse checkout at a pinned SHA keeps the reader byte-identical to upstream with no copy to re-sync on schema bumps; the pin makes reader updates an explicit SHA bump. `SUPPORTED_SCHEMA_MAJOR` in the consumer and the pinned SHA must be bumped together on a schema-major change.
- **Record on pass only** — a green run is the sole conclusive "tested." Any failure (transient infra *or* a real build/test break) leaves the version unrecorded so the next poll re-tests it, rather than masking it until the 7-day cache eviction. Retry is bounded: `resolve()` always prefers the newest ready build, so a genuinely broken nightly is superseded within the morning poll window, not retried hourly.
- **Matrix = full nightly minus preinstalled images** — 4 multi-arch distros × {gfx1100, gfx1151} × {whl, tarball} = 16 legs. The version-pinned stable image is excluded (via `configure_ci.py --exclude-preinstalled`) because its ROCm version is baked in and unrelated to the nightly we pin.
- **Poll window** — every 2h from 01:30–11:30 UTC. The Linux ROCm build starts ~00:02 and goes green ~02:24–03:09 on a normal night, so 01:30 sees it in progress (cheap skip), 03:30 usually catches it green, and the morning tail absorbs publish lag. A morning finish drops only the rare evening re-run, which self-heals when the next nightly supersedes it.
- **Cache eviction** — the 7-day marker eviction is accepted (a stale week means one harmless re-run).
- **`repository_dispatch`** (Quartz pushing on each build, removing polling+dedup) is deferred as a follow-up.

## Test plan

- [x] `consume_status.py` unit tests pass (`PYTHONPATH=<quartz>/scripts/consumer python3 -m unittest discover -s .github/quartz/tests`)
- [x] Consumer resolves the live Quartz `latest.json` (build=success) and emits the pinned version + install URLs
- [x] `configure_ci.py --exclude-preinstalled` drops the preinstalled image and keeps `distros`/`distro_map` in sync
- [x] Both workflow YAMLs parse
- [x] Triggered run: gate resolves and the matrix pins the exact nightly (observed — gate resolved `10.1.0a20260902`; whl leg installed `rocm==10.1.0a20260902`)
- [ ] `record` writes the dedup marker on a **green** run (blocked: current nightly fails to build due to TheRock `libhipcxx`/`libomp` packaging, so no green run yet)
- [ ] Second immediate run shows the **skip** (already-tested) report


.github/workflows/quartz_test.yml
```diff
@@ -0,0 +1,192 @@
...
+    # Quartz republishes latest.json. The 2h gap between polls is itself the retry --
+    # a build still in progress at one poll is almost always done by the next, and it
+    # rides out transient Quartz outages. A morning finish drops only the rare evening
+    # re-run, which self-heals when the next nightly supersedes it. (Quartz's own
+    # example uses 17 5-19; we start earlier and finish sooner to cut latency, since
+    # our build-only gate is ready well before the full release. See ROCm/Quartz PR #32.)
+    - cron: "30 1-11/2 * * *"
+  # TEMPORARY (REVERT BEFORE MERGE): run on PRs touching this workflow so it can
+  # be validated end-to-end without waiting for the schedule (which only fires on
+  # the default branch). Remove this pull_request trigger before merging.
```
zichguan-amd: This can be removed

zichguan-amd (approved):
LGTM thanks!

PR #514, opened 2026-09-03
fix(rocProfiler-SDK): buffered-tracing race + thread_trace API sync

zichguan-amd:
## Summary

Two related fixes bringing the rocProfiler-SDK examples back in line with the upstream samples (ROCm/rocm-systems, `projects/rocprofiler-sdk/samples`), which had diverged.

### 1. Fix buffered-tracing data race + sync examples with upstream
- **api_buffered_tracing**: guard the shared `call_stack_t` (`tool_data`) vector with a mutex. It is written concurrently from the async buffer-flush thread and the `thread_precreate`/`thread_postcreate` callbacks with no synchronization, so overlapping reallocations corrupt the heap and abort with "double free or corruption" on the first kernel launch. This was the root cause of the timing-dependent CI aborts of `rocprofiler-sdk_api_buffered_tracing`. Verified on a repro host: original crashed 5/5, fixed passed 10/10.
- **Common/rocprofiler_utils.hpp**: make `ROCPROFILER_CHECK`/`ROCPROFILER_CALL` use a `__LINE__`-uniqued internal variable (`ROCPROFILER_VARIABLE(CHECKSTATUS, __LINE__)`), matching upstream's `defines.hpp` and the existing `ROCPROFILER_WARN`. The previous plain `status` collided with a caller variable named `status`.
- **pc_sampling**: check the real status from `rocprofiler_destroy_buffer` instead of the constant `ROCPROFILER_STATUS_SUCCESS` (which made the check a no-op); fix "availabe" typo.
- **api_callback_tracing, openmp_target**: wrap `rocprofiler_is_initialized()` in `ROCPROFILER_CALL` so a failed query is not silently ignored (relies on the macro-hygiene fix above).

### 2. Update thread_trace to current SDK thread-trace API
- The experimental thread-trace shader-data callback contract changed between ROCm releases: `rocprofiler_thread_trace_shader_data_callback_t` now takes a single `rocprofiler_thread_trace_shader_data_t` struct. This struct-based form is available since ROCm 10; the pinned 7.14 stable image still ships the older multi-argument callback (`agent, shader_engine_id, data, data_size, flags, userdata`) and has no such struct, so the updated example does not compile there.
- `shader_data_callback`: take the struct; decode from `shader_data.data + read_offset` for `shader_data.data_size - read_offset` bytes.
- `tool_codeobj_tracing_callback`: only handle `ROCPROFILER_CALLBACK_PHASE_LOAD`.
- Brace-init the thread-trace parameter union values.
- **Enable `thread_trace` in the suite CMake** (`add_subdirectory`). It was previously commented out ("Disabled until feature is released"), which is why the API break escaped CI. It now builds on the nightly TheRock channel (which carries the new API).
- **Skip `thread_trace` on the pinned 7.14 stable image** via a stable-channel build entry in the skip manifest (`.github/build_tools/skip_manifest.py`), since that image predates the new struct API. Nightly builds are unaffected and still compile it.

## Test plan
- [x] Full-suite cmake build with `-DROCM_EXAMPLES_ENABLE_OPENMP=ON` — configure + build exit 0, all targets including `rocprofiler-sdk_thread_trace` produced.
- [x] Skip manifest: `--channel stable` lists `Libraries/rocProfiler-SDK/thread_trace` in `skip_build.txt` and the CMake override drops it (`SkipExamples: skipping ...`); `--channel nightly` does not skip it.
- [x] api_buffered_tracing race repro: 5/5 crash before, 10/10 pass after.
- [ ] CI green: nightly builds thread_trace, 7.14 stable skips it.

### DIFF
--- a/.github/build_tools/skip_manifest.py
+++ b/.github/build_tools/skip_manifest.py
@@ -140,6 +140,19 @@ SKIP_MANIFEST = [
         "channels": ["stable"],
         "reason": "hip_scan.h not present in the pinned 7.14 stable image",
     },
+    # --- Stable-only build skip ------------
+    # The experimental thread-trace shader-data callback was changed to take a
+    # single rocprofiler_thread_trace_shader_data_t struct; the pinned 7.14 stable
+    # image still ships the older multi-arg callback and has no such struct, so the
+    # updated example does not compile there. Nightly TheRock builds carry the new
+    # API and build it, so scope this skip to the stable channel only.
+    {
+        "ctest": None,
+        "path": "Libraries/rocProfiler-SDK/thread_trace",
+        "scope": ["build"],
+        "channels": ["stable"],
+        "reason": "rocprofiler_thread_trace_shader_data_t (new single-struct shader-data callback) absent from the pinned 7.14 stable image",
+    },
     # --- hipThreads: nightly whl build skip (whole tree) ------------------
     # The TheRock nightly whl ships hipthreads headers + libhipthreads.a but an
     # empty include/libhipcxx, so the libhipcxx headers the examples include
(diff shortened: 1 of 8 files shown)


mapatel-amd (approved):
LGTM

PR #517, opened 2026-09-09
ci(quartz): classify infra vs build/test failures; retry until infra is healthy

zichguan-amd:
## Motivation

The Quartz nightly test marks a whole run `failure` without distinguishing a
transient infra problem (an unpublished wheel, a tarball 404, a runner hiccup)
from a genuine build/test break. Both cases are treated the same: the version is
never recorded, so the full self-hosted GPU matrix re-runs every 2h poll — even
when the failure is a real, permanent break that retrying can never fix.

Last night (ROCm 10.1.0a20260909) is a textbook case: every `whl-multi-arch`
matrix job died at install with `No matching distribution found for rocm==10.1.0a20260909`
(the wheel simply wasn't published yet — pure infra), while a couple of matrix
jobs that installed fine hit genuine breaks (`libavcodec.so.58` missing, `rocgdb`
ctest failures). Today's workflow can't tell these apart.

## Technical Details

Each matrix job now self-reports a category (`infra` / `bug` / `success`) as an
uploaded artifact, and a `collect` job aggregates them to drive the
record-vs-retry decision:

- **Per-job classification** (`build-rocm-examples-reusable.yml`): an
  `Environment status check` step flags checkout/install/env-setup failures — and
  the wheel sanity check — as transient. Build and test steps are gated with
  `!cancelled()` so the CMake and Make chains fail-fast **independently** rather
  than cascading. `Classify outcome` then emits `infra` (env/install broke, or an
  artifact upload failed), `bug` (a build/test step failed), or `success`,
  uploaded as `status-<distro>-<gpu>-<install_method>`. The artifact body carries
  the bare category on its first line (so `collect` can match it with `grep -qx`)
  followed by the resolved `rocm_version` and `install_url`.
- **Aggregation is infra-dominates** (`quartz_test.yml` `collect` job): if **any**
  matrix job hit an infra failure, the whole version is `infra` and is left
  unrecorded so the next poll retries it — i.e. **retry until infra is healthy**.
  Only once every matrix job's environment came up does a genuine build/test
  break settle it as `bug`. `collect` also reports the verdict to the run summary,
  including the `infra` (will-be-retried) case that no other job surfaces.
- **`record`** writes the dedup marker on a clean pass **or** a `bug` (retrying
  won't fix broken examples — mark tested, stop re-running, surface a warning).
  An `infra` verdict stays unrecorded. It is guarded with `always()` so a failed
  or skipped `collect` cannot skip recording an all-green matrix. Retry is
  bounded: a version is only revisited while it stays Quartz's latest-ready build,
  and the next nightly supersedes it.

Also includes naming cleanup of the per-job status plumbing
(`prep`→`environment-status`, `mk-build`/`mk-test`→`make-build`/`make-test`,
`cell-category`/`category-*`→`job-status`/`status-*`), gates `Run tests` on
`cmake-build` instead of `cmake-install`, corrects a stale comment that
mis-attributed the per-job classification to the `collect` job (it is done by the
local `Classify outcome` step), and pins `actions/checkout` and
`actions/upload-artifact` to commit SHAs. (Pinning the remaining tag-pinned
workflows and adding a `check-gha-pinning` pre-commit hook to enforce it is left
to a follow-up PR.)

## Test Plan

- YAML + actionlint validated for both workflows.
- Logic traced against last night's real matrix (run 34300488752): whl/tarball
  install failures → `infra` (dominates) → version stays unrecorded and retries;
  once installs are healthy, the `libavcodec`/`rocgdb` breaks would classify
  `bug` → record + warn.

## Test Result

Full end-to-end behavior is exercised by the scheduled nightly against the next
ready Quartz build; not runnable ad hoc here.

- [ ] Yes

- [ ] Yes

## Submission Checklist

### DIFF
--- a/.github/workflows/build-rocm-examples-reusable.yml
+++ b/.github/workflows/build-rocm-examples-reusable.yml
@@ -58,10 +58,12 @@ jobs:
         gpu_config: ${{ fromJson(inputs.gpu_configs) }}
     steps:
       - name: Checkout repository
+        id: checkout
         uses: actions/checkout@v4
 
       - name: Install ROCm (whl-multi-arch)
         if: ${{ matrix.install_method == 'whl-multi-arch' }}
+        id: install-whl
         run: |
           echo "## Target: ${{ matrix.gpu_config.gpu_target }} on ${{ fromJson(inputs.distro_map)[inputs.distro].label }}" >> $GITHUB_STEP_SUMMARY
           pip install --no-cache-dir --upgrade pip
@@ -127,6 +129,7 @@ jobs:
 
       - name: Setup environment variables (tarball-multi-arch)
         if: ${{ matrix.install_method == 'tarball-multi-arch' }}
+        id: setup-env-tarball
         run: |
           ROCM_PATH=/therock-tarball/install
           echo "ROCM_PATH=${ROCM_PATH}" >> $GITHUB_ENV
@@ -154,7 +157,29 @@ jobs:
           echo "ENABLE_CK=OFF" >> $GITHUB_ENV
           echo "HIPCC_COMPILE_FLAGS_APPEND=--offload-arch=${{ matrix.gpu_config.gpu_target }}" >> $GITHUB_ENV
 
+      - name: Prep status
+        id: prep
+        if: always()
+        run: |
+          # "Prep" = everything needed before our builds: checkout, ROCm install, env
+          # setup. A failure here is transient infra (network/index/runner), not a code
+          # bug -- classified infra so the version is retried, not recorded broken.
+          ok=true
+          for oc in "${{ steps.checkout.outcome }}" \
+                    "${{ steps.install-whl.outcome }}" \
+                    "${{ steps.whl-multiarch-version.outcome }}" \
+                    "${{ steps.install-tarball-multi-arch.outcome }}" \
+                    "${{ steps.setup-env-tarball.outcome }}" \
+                    "${{ steps.preinstalled.outcome }}"; do
+            [ "$oc" = "failure" ] && ok=false
+          done
+          echo "ok=$ok" >> "$GITHUB_OUTPUT"
+
+      # The collect job in quartz_test.yml classifies each cell from these build/test
+      # step ids/outcomes; keep the ids stable when renaming steps.
       - name: Generate skip lists
+        if: ${{ !cancelled() && steps.prep.outputs.ok == 'true' }}
+        id: skip-lists
         run: |
           # The preinstalled distro is the pinned rocm:7.14 image, i.e. the "stable"
           # channel; every other install method pulls a TheRock "nightly" build.
@@ -170,17 +195,21 @@ jobs:
             --install-method "${{ matrix.install_method }}"
 
       - name: Makefile build
+        if: ${{ !cancelled() && steps.prep.outputs.ok == 'true' && steps.skip-lists.outcome == 'success' }}
+        id: mk-build
         run: |
           ./Scripts/configure.sh --rocm-path="${ROCM_PATH}"
           make -j HIP_ARCHITECTURES="${{ matrix.gpu_config.gpu_target }}" SKIP_FROM_BUILD="$SKIP_FROM_BUILD"
 
       - name: CMake configure
-        if: ${{ !cancelled() }}
+        if: ${{ !cancelled() && steps.prep.outputs.ok == 'true' && steps.skip-lists.outcome == 'success' }}
+        id: cmake-configure
         run: |
           cmake -S . -B build -DCMAKE_HIP_ARCHITECTURES="${{ matrix.gpu_config.gpu_target }}" -DCMAKE_BUILD_RPATH="${ROCM_PATH}/lib" -DROCM_EXAMPLES_ENABLE_OPENMP="${ENABLE_OPENMP}" -DROCM_EXAMPLES_ENABLE_CK="${ENABLE_CK}" -DCMAKE_PROJECT_INCLUDE_BEFORE="${GITHUB_WORKSPACE}/Common/SkipExamples.cmake" 2> >(tee cmake_error.log >&2)
 
       - name: CMake configure error summary
         if: ${{ !cancelled() }}
+        continue-on-error: true
         run: |
           if [ ! -f cmake_error.log ]; then
             exit 0
@@ -211,17 +240,20 @@ jobs:
           fi
 
       - name: CMake build
-        if: ${{ !cancelled() }}
+        if: ${{ !cancelled() && steps.cmake-configure.outcome == 'success' }}
+        id: cmake-build
         run: |
           cmake --build build -j
 
       - name: CMake install
-        if: ${{ !cancelled() }}
+        if: ${{ !cancelled() && steps.cmake-build.outcome == 'success' }}
+        id: cmake-install
         run: |
           cmake --install build --prefix install
 
       - name: Collect Makefile build artifacts
         if: ${{ !cancelled() }}
+        continue-on-error: true
         run: |
           mkdir -p /tmp/makefile_build
           find . -name Makefile -not -path './build/*' | while IFS= read -r makefile; do
@@ -235,6 +267,7 @@ jobs:
 
       - name: Upload CMake build artifacts
         if: ${{ !cancelled() }}
+        continue-on-error: true
         uses: actions/upload-artifact@v4
         with:
           name: rocm-examples-build-${{ inputs.distro }}-${{ matrix.gpu_config.gpu_target }}-${{ steps.whl-multiarch-version.outputs.rocm_version || steps.install-tarball-multi-arch.outputs.rocm_version || steps.preinstalled.outputs.rocm_version }}-${{ matrix.install_method }}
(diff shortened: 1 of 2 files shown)


.github/workflows/quartz_test.yml
```diff
@@ -155,7 +190,11 @@ jobs:
 
       - name: Summary
         run: |
-          echo "Recorded ROCm ${{ needs.gate.outputs.rocm_version }} (${{ needs.gate.outputs.build_date }}) as tested (passed)." >> "$GITHUB_STEP_SUMMARY"
+          if [ "${{ needs.test.result }}" = "success" ]; then
+            echo "Recorded ROCm ${{ needs.gate.outputs.rocm_version }} (${{ needs.gate.outputs.build_date }}) as tested (passed)." >> "$GITHUB_STEP_SUMMARY"
+          else
```
HereThereBeDragons: arent you missing here the 3rd option: infra == retry?

.github/workflows/build-rocm-examples-reusable.yml
```diff
@@ -170,17 +196,21 @@ jobs:
             --install-method "${{ matrix.install_method }}"
 
       - name: Makefile build
+        if: ${{ !cancelled() && steps.environment-status.outputs.ok == 'true' && steps.skip-lists.outcome == 'success' }}
```
HereThereBeDragons:
```suggestion
        if: ${{ !cancelled() && steps.skip-lists.outcome == 'success' }}
```
dont think we need the other one - same for all the following ones

.github/workflows/build-rocm-examples-reusable.yml
```diff
@@ -298,23 +335,55 @@ jobs:
...
+          elif [ "${{ steps.skip-lists.outcome }}" = "failure" ] \
+            || [ "${{ steps.cmake-configure.outcome }}" = "failure" ] \
+            || [ "${{ steps.cmake-build.outcome }}" = "failure" ] \
+            || [ "${{ steps.cmake-install.outcome }}" = "failure" ] \
+            || [ "${{ steps.ctest.outcome }}" = "failure" ] \
+            || [ "${{ steps.make-build.outcome }}" = "failure" ] \
+            || [ "${{ steps.make-test.outcome }}" = "failure" ]; then
+            c=bug
+          else
+            c=success
```
HereThereBeDragons, line 387: now we would get success even if the upload failed. is that something that is ok or where you want a rerun? in theory i guess this would count as "infra" problem?

.github/workflows/build-rocm-examples-reusable.yml
```diff
@@ -298,23 +335,55 @@ jobs:
...
+          fi
+          echo "$c" > job-status.txt
```
HereThereBeDragons: wouldnt you also want then to add the rocm version etc?

.github/workflows/quartz_test.yml
HereThereBeDragons, line 233: want to also report on `needs.collect.outputs.category`? that it will be retried?

.github/workflows/build-rocm-examples-reusable.yml
```diff
@@ -298,23 +335,55 @@ jobs:
...
+          echo "Job status: $c" >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Upload job status
+        if: always()
+        uses: actions/upload-artifact@v4
```
HereThereBeDragons: pin to sha

HereThereBeDragons: really recommend to use pre-commit to catch those. you can just copy it from therock

.github/workflows/build-rocm-examples-reusable.yml
HereThereBeDragons, line 88: do we need this in the environment check?

.github/workflows/quartz_test.yml
```diff
@@ -128,17 +129,51 @@ jobs:
...
-    # Record ONLY on a green test run. Any failure -- transient infra (wheel-index
-    # outage, tarball 404, runner hiccup) OR a real build/test break -- leaves the
-    # version unrecorded so the next poll re-tests it, instead of masking it until
-    # the 7-day cache eviction. A failed run is never a conclusive "tested": we may
-    # simply have been unable to install. Retry is bounded -- a version is revisited
-    # only while it stays Quartz's latest-ready build, and the next nightly (which
-    # resolve() always prefers) supersedes it -- so a genuinely broken nightly costs
-    # at most the handful of polls in one morning window, not an hourly loop.
-    if: needs.test.result == 'success'
+    # Read each cell's self-reported category and retry until infra is healthy: any
```
HereThereBeDragons: what is a cell in this context?

.github/workflows/quartz_test.yml
```diff
@@ -128,17 +129,51 @@ jobs:
...
+          # all -> infra, so we retry rather than record a version we have no signal for.
+          if   grep -rqx infra   status 2>/dev/null; then category=infra
+          elif grep -rqx bug     status 2>/dev/null; then category=bug
+          elif grep -rqx success status 2>/dev/null; then category=success
+          else                                            category=infra
+          fi
+          echo "category=$category" >> "$GITHUB_OUTPUT"
+
+  record:
+    needs: [gate, test, collect]
```
HereThereBeDragons, line 180: one comment from claude to think about:

quartz_test.yml:168 (record) — record needs collect; since its if lacks always(), a failed collect skips record even on all-green. The test.result == 'success' clause can't rescue it. Self-healing via retry, but the condition implies robustness it doesn't have.

.github/workflows/build-rocm-examples-reusable.yml
zichguan-amd: Good idea, this should belong to the env check

.github/workflows/build-rocm-examples-reusable.yml
zichguan-amd: Upload doesn't matter that much, but yes in theory this should belong to infra

.github/workflows/quartz_test.yml
zichguan-amd: It's a matrix job, will rephrase it

.github/workflows/build-rocm-examples-reusable.yml
zichguan-amd: pre-commit and other workflow SHA pin will be a follow up PR, pinned all actions in current PR scope.

.github/workflows/build-rocm-examples-reusable.yml
zichguan-amd: Added rocm version and install url. Other infos can be infered from file name after upload

.github/workflows/build-rocm-examples-reusable.yml
zichguan-amd: explicitly added those to infra

.github/workflows/quartz_test.yml
zichguan-amd: Added always to assure the success case works independently of the collect step

.github/workflows/build-rocm-examples-reusable.yml
```diff
@@ -374,12 +386,20 @@ jobs:
           else
             c=success
           fi
-          echo "$c" > job-status.txt
+          # First line is the bare category (collect aggregates it with `grep -qx`);
+          # the metadata lines below it are inert to that match.
+          rocm_version="${{ steps.whl-multiarch-version.outputs.rocm_version || steps.install-tarball-multi-arch.outputs.rocm_version || steps.preinstalled.outputs.rocm_version }}"
```
HereThereBeDragons, line 391: why dont use `inputs.`?

.github/workflows/quartz_test.yml
```diff
@@ -165,6 +165,17 @@ jobs:
           fi
           echo "category=$category" >> "$GITHUB_OUTPUT"
 
+      - name: Report verdict
+        run: |
+          # record itself reports the success/bug cases; the infra verdict is not
+          # recorded by any downstream job, so surface here that it will be retried.
+          ver="ROCm ${{ needs.gate.outputs.rocm_version }} (${{ needs.gate.outputs.build_date }})"
```
HereThereBeDragons, line 172: again why not `inputs.`?

HereThereBeDragons (commented):
lgtm other than the comments below

.github/workflows/build-rocm-examples-reusable.yml
zichguan-amd: `rocm_version` is not a required input for workflows other than calling from quartz, so we need to handle the case where it's getting latest at runtime or parse a preinstalled stable version.

.github/workflows/quartz_test.yml
zichguan-amd: There's no `rocm_version` in inputs for this workflow, would be getting from quartz at `consume`'s outputs or `gate`'s outputs (which is what I used)

.github/workflows/build-rocm-examples-reusable.yml
HereThereBeDragons: ok - other option would be to enforce that you give a version?

.github/workflows/quartz_test.yml
HereThereBeDragons: ah i missed that this was already the next workflow (:

.github/workflows/build-rocm-examples-reusable.yml
zichguan-amd: Could reserve a string for latest/preinstalled, but I think the current version discovery is pretty reliable already. It's getting it from `rocm-skd version` or `/opt/rocm/core/.info/version`, or parsed version string from url.

PR #522, opened 2026-09-15
ci: resolve gitleaks + zizmor + trivy security scanner findings

zichguan-amd:
## Summary
Resolves outstanding HIGH findings reported by the security scanners added in this branch (zizmor, trivy, gitleaks), targeting `users/cgoea/security`.

- **zizmor template-injection** — move workflow `inputs.*` into step-level `env:` and reference `$VAR` in `run:` blocks:
  - `build-rocm-examples-reusable.yml`: `install-whl`, `install-tarball-multi-arch`, `skip-lists`
  - `quartz_test.yml`: `dedup` step (`github.event.inputs.force`)
- **zizmor unpinned-images**:
  - `build_libraries_cuda.yml`: pin the CUDA container to its current `@sha256` digest (tag kept as a comment)
  - `build-rocm-examples-reusable.yml:45`: suppress the runtime-computed `fromJson(inputs.distro_map)[...]` image, which cannot be statically digest-pinned
- **zizmor excessive-permissions** (`build_docker.yml`): drop workflow-level `packages: write` (granted to every job) and scope it to the `build` job that pushes to GHCR; `find-dockerfiles` stays read-only
- **zizmor linting.yml**: suppress `unpinned-uses` (`rocm-docs-core@develop`) and `template-injection` (git-clang-format base-ref) — both intentional
- **gitleaks generic-api-key** (`gitleaks.toml`): rule-scoped allowlist for Visual Studio `.sln` `Project("{GUID}") = ...` declarations (3 false positives). A finding is dropped only when it is both in a `.sln` file AND on a `Project()` line (`condition = "AND"`), so real keys and other rules are unaffected.
- **trivy DS-0017** (`ubuntu-24.04-rocm.Dockerfile`): pair `apt update` with `apt install` in the same `RUN` layer
- **trivy DS-0002 / DS-0029**: suppress two non-applicable HIGH Dockerfile findings.
  - DS-0002 (non-root USER) — these images are GitHub Actions job-containers that must run as root; a non-root USER cannot write `GITHUB_WORKSPACE`, verified empirically. Suppressed via a repo-root `.trivyignore`, as the finding has no line attribution and cannot be scoped inline.
  - DS-0029 (`--no-install-recommends` on the `rocm` meta-package, whose recommends pull in required ROCm subcomponents) — suppressed via an inline `# trivy:ignore:AVD-DS-0029` comment scoped to `Dockerfiles/hip-libraries-cuda-ubuntu.Dockerfile`, its single justified occurrence.

## Test plan
- [x] `zizmor --offline .github/workflows/` reports 0 HIGH across the tree
- [x] `gitleaks detect` reports 0 leaks; negative test confirms a real key planted in a `.sln` is still caught
- [x] `trivy config Dockerfiles/` reports 0 HIGH/CRITICAL
- [ ] CI green on `users/cgoea/security`


zichguan-amd:
CI only run on changed files compared to base, local full scan still shows red on zizmor and trivy

zichguan-amd:
local full scan is now clean

cgoea (approved):
LGTM.

